PR #17, opened 2018-05-28
Add optional 'MAGICC_EXECUTABLE' environment parameter

lewisjared:
If environment variable 'MAGICC_EXECUTABLE' is set, this binary is used run instead of the included version of magicc.

### DIFF
--- a/pymagicc/__init__.py
+++ b/pymagicc/__init__.py
@@ -29,13 +29,20 @@ __version__ = get_versions()["version"]
 del get_versions
 
 
+def get_paths():
+    default_executable = os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        "MAGICC6/MAGICC6_4Download/magicc6.exe"
+    )
+    
+    executable = os.environ.get('MAGICC_EXECUTABLE', default_executable)
+    return os.path.dirname(executable), os.path.basename(executable)
+
+
+_magiccpath, _magiccbinary = get_paths()
+
 _WINDOWS = platform.system() == "Windows"
 
-_magiccpath = os.path.join(
-    os.path.dirname(os.path.abspath(__file__)),
-    "MAGICC6/MAGICC6_4Download"
-)
-_magiccbinary = "magicc6.exe"
 
 if not _WINDOWS:
     wine_installed = subprocess.call("type wine", shell=True,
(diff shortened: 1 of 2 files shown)


rgieseke:
Hi @lewisjared - many thanks! I'm super happy to make any changes in Pymagicc that make it easier to run more tests for MAGICC and enable Continuous Integration (just spoke with @znicholls)!

I was just wondering, why do you need this as an environment variable? I'm not in principle opposed to this change, just worried that it might lead to hard-to-debug issues in the future with e.g. Windows users in the future (I would have to look up on where this might be set-up on Windows), in a script it would always be explicit.

Is your main use case being able to change the whole MAGICC package (binary, config, etc.) in one go?

lewisjared:
Hi @rgieseke 

Correct, the use case I intended was to be able to specify an entire new MAGICC package including configuration, either for local development or automated testing. The location of the new package is somewhat dynamic so it may change.

I decided to use ENV variables as they provide the cleanest mechanism for dynamically specifying the location of MAGICC. It was also the simplest to implement!

An example of how I envision running a script which uses a custom MAGICC package using bash:
`MAGICC_EXECUTABLE=/tmp/magicc/magicc.exe python run_tests.py`
or if many scripts are to be run
```
#!/bin/bash
export MAGICC_EXECUTABLE=/tmp/magicc/magicc.exe
python run_tests.py
python generate_plots.py
```
Any other python interpreters using pymagicc would not be impacted by this change and would default to the embedded version of MAGICC.

Other options that would work include:
- Config file
This would enable more customisation of pymagicc but would require a bit more thought about what goes in this file, what format to use and where to put it. This would add some additional complexity when running dynamic tests as new config files need to be written potentially impacting the global usage of pymagicc in other python interpreters.
- Function parameters
More explicit. Requires quite a few changes to the codebase 

Often there is a hierarchy of configuration variables which override each other. Something like: function parameters > ENV variables > config file > default config file. It would be great to also have configuration files in future as they are well suited to static variables, but ENV variables is a temporary stop-gap for now.

rgieseke:
Alright, thanks for the explanation! 
I'll just rename `get_paths` to `_get_magicc_paths` ...

znicholls:
Also from me, thanks for the explanation. I look forward to seeing how this hierarchy of configs can be used well and how we test it all over time :)

PR #18, opened 2018-05-30
Add version specific parameters depending on the version of MAGICC

lewisjared:
This is a version agnostic implementation of @znicholls version of a pymagicc which works with MAGICC7.

Adds a MAGICC_VERSION environment variable (default="6") which is used to specify the target version of MAGICC.

I have refactored out the version specific parameters used within pymagicc into the `compat` module. Additionally, the logic for determining where the config/output has gotten a little more complex as MAGICC7 will have different run and output folders.

All tests on MAGICC6 are still working, as well as a majority of tests under MAGICC7. The run_x tests are currently failing as the global surface temperatures are slightly different to the stored values from MAGICC6. 

@znicholls test_default_config also fails as the CORE_CLIMATESENSITIVITY in PARAMETERS.out is 2.5. I used the snippet for MAGICC_USER.CFG. Any ideas?


znicholls:
Wicked looks awesome! I think the `CORE_CLIMATESENSITIVITY` is being overwritten by something else. Maybe XTRAPARAS or something. I'll have a look.

rgieseke:
Great! I haven't tested this yet, but one question:

I'm a bit reluctant to introduce another environment variable, could there be another way to detect the MAGICC version? If `MAGICC_EXECUTABLE` is set one could look in this folder for config file key or something that defines the version for example.

rgieseke:
I'm also quite happy to include a copy of `RCP3PD.SCEN` as `RCP26.SCEN` or rename this file if this makes future compatibility easier. I don't thinkt there are many  users yet who would rely on it being named RCP3PD :-)

Another thing, MAGICC6 can also have separate input and out directories I think, so I'm not sure whether this should be set as such in compat. We could also change this in the embedded MAGICC6 version if it's easier for compatibility.

rgieseke:
We should probably tag the tests which can only succeed with the embedded version

`@pytest.mark.skipif`

https://docs.pytest.org/en/2.9.0/skipping.html#id1

znicholls:
- I just realised, we can tell which MAGICC version it is by looking in `MAGCFG_DEFAULTALL.CFG`, if there's a `FILE_EMISSIONSCENARIO` flag then it's `MAGICC6`, if there's a `FILE_EMISSCEN_2` flag then it's `MAGICC7`, if both fail then `MAGICC` will blow up if you try to run it so throw an error
- moving to make `pymagicc`'s `MAGICC6` set up mirror that of `MAGICC7` will certainly make things easier. I'm almost certain that you can just change the settings in `MAGCFG_DEFAULTALL.CFG` to get this functionality
- good idea re `pytest.mark.skipif` Robert!

znicholls:
Jared I've just had a look re that ECS issue. You're going to have to issues with the current set up in `magicc/magicc`. At the moment `MAGCFG_USER.CFG` is set up like:

```FORTRAN
  FILE_TUNINGMODEL_1 =  "C4MIP_UVIC",  
  FILE_TUNINGMODEL_2 =  "FULLTUNE_MEDIUM_CMIP3_ECS2_5",  
  FILE_TUNINGMODEL_3 =  "USER",  
  FILE_TUNINGMODEL_4 =  "USER",  
  FILE_TUNINGMODEL_5 =  "USER",  
  FILE_TUNINGMODEL_6 =  "USER",  
  FILE_TUNINGMODEL_7 =  "USER",  
  FILE_TUNINGMODEL_8 =  "USER",  
  FILE_TUNINGMODEL_9 =  "USER",  
  FILE_TUNINGMODEL_10 =  "XTRAPARAS_THISRUN",  
```

`FILE_TUNINGMODEL_1 =  "C4MIP_UVIC"` if fine as that's only carbon cycle related parameters. `FILE_TUNINGMODEL_2 =  "FULLTUNE_MEDIUM_CMIP3_ECS2_5"` is the problem as it contains the line, `CORE_CLIMATESENSITIVITY =  2.5` which is over-writing the ECS of 3 which you're intending on having. `FILE_TUNINGMODEL_10 =  "XTRAPARAS_THISRUN"` is also a problem as it means that your setup is locked from `pyamgicc` (for an explanation see the text I put in the `lazy-pymagicc-testing` notebook under `Round Two`). I'll put the setup in a couple of branches in `magicc/magicc` for you to have a look at.

rgieseke:
I've renamed `RCP3PD` to `RCP26`.

lewisjared:
If RCP6.SCEN was renamed to RCP60.SCEN that would also be great! Then I can remove all mention of scenarios from `pymagicc.compat`

lewisjared:
I removed the dependency on the MAGICC_EXECUTABLE environment variable, instead, the MAGICC version is determined by looking at MAGCFG_USER.CFG. The number of version specific was also reduced by making the embedded version of MAGICC6 mimic the default folder structure seen in MAGICC7

lewisjared:
Also, this change will impact end users who specify an output folder as the output will now be in an out subfolder under the directory they specified. Not a huge change, but it will break somethings :D.

We could cherrypick all, but f66c60c from this PR to keep the same folder location for MAGICC6 and include that commit at a later date before a major version bump.

rgieseke:
This looks good, but I'm wondering if it could be done without caching the version numbers and a separate compat module. 

Could we maybe just check whether a file like `MAGTUNE_XTRAPARAS_THISRUN.CFG` exists? Maybe a different one would be better, @znicholls ?

I don't think the `out` location is a problem. 

Will there be more params that will need adjustment? Otherwise I'm wondering if an approach to check just once for the version used and define constants is possible.

Something like:
```python
MAGICC7 = True  # if file exists or something

# ...

skiprows=21 if MAGICC7 else 19
```
etc.

znicholls:
The only thing I can think that would be a safe check is: looking in MAGCFG_DEFAULTALL.CFG, if there's a FILE_EMISSIONSCENARIO flag then it's MAGICC6, if there's a FILE_EMISSCEN_2 flag then it's MAGICC7, if both fail then MAGICC will blow up if you try to run it so throw an error

I'm not yet sure how many parameters will need adjustments. At the moment there are very few parameters you would change with `pymagicc` (could is a different question) I suspect if you did a diff between a MAGICC6 and MAGICC7 CFG it would be closer to 20 than 5 so I think the `compat` module makes sense...

In general I am in favour of more, smaller methods/functions (small to the point where doc strings don't make sense because it's obvious what everything does) rather than avoiding adding extra modules and caches etc. Trying to keep everything in one place makes me nervous because it can quickly get too big to get your head around (personally I find anything with more than 20 lines tricky to understand and would ideally split up).

rgieseke:
Fully agree on the global variable issue, now that we have agreed on ideally supporting both versions at the same time, this doesn't make sense anymore.

PR #19, opened 2018-05-31
Lower level API

lewisjared:
For testing, it would be useful to have a lower level api than the `run` function. This would enable more control over the model configuration between the creation of new folders and the running of the model. I think that it would also be a useful addition to the public API. This also simplifies the `run` function.

Encapsulating this functionality in a class is cleaner than a collection of functions as there is some state that needs to be tracked between creating the folder hierarchy and running the model (folder paths, is it a temp directory, results, parameters).

```
from pymagicc.run import ModelRun

model_run = ModelRun()
# Create/edit configuration
results = model_run.run()
parameters_out = model.config
```

This branch is off magicc7-compat so only the last two commits are of interest.


znicholls:
I think we're now asking a question about how we want pymagicc` to look and behave. Thoughts on the below would be great.

At the moment `pymagicc` is a simple wrapper which does the basic jobs for running MAGICC i.e. it writes your `SCEN` file for you and will set the `NML_YEARS.CFG` file too. It can also write your `CFG` file assuming that you're using `pymagicc`'s binary and not anything else where the config might not get through. You can read through the code in 20 minutes or so and see what is going on. The benefits of this are simplicity and accessibility. The expense is limited config ability, slow to run (cause of this entire directory copying thing) and if we have to maintain such simplicity we're kind of locked in how far we can change things. 

My ideal would be that `pymagicc` offers a number of things. At the highest level, this simple API, where a user can give it a pandas dataframe with scenario data and a few configs, run magicc and get results and configs back in dataframes. However I think we should also be able to offer people the ability to do probabilistic runs for a number of scenarios, with somewhat complex `CFG` file input etc. if they wish and take the time to learn how to set things up. Offering such control necessarily requires a more complex backend.

I think the simple API is something we should maintain always (and can ensure this is so with the tests). However I think having a simple API at a high level should not preclude us from making things more complex at a lower level. If we have the tests, we can guarantee our user experience whilst also giving us the freedom to make the back end also support much more complexity. If we're going to use `pymagicc` for testing `MAGICC` then I think we have to support extra complexity and so the backend has to become more complex. Ultimately we have to make a choice about what level of abstraction we want to go to and I would rather that we offer the API, which shows you outputs easily so you can always see what's going on, but if you want to contribute to the code base you're going to have to know something about `python` so at some point we're going to use more sophisticated tools e.g. classes, lots of small methods and multiple modules. 

Ultimately this is my ideal, if maintaining simplicity is the most important thing for `pymagicc` then I guess that's the most important thing and we go to plan b.

rgieseke:
Alright, let's do this! These are good points, @znicholls 
I think we should have issues with concrete use-cases and maybe split this into smaller PRs to make it easier to review.

@lewisjared We're trying to abstract away the "copy" of the MAGICC package, right? We want to be able to create a copy, modify config or inputs, run and re-run ...?
Should the object, the abstraction then be "Package" or some better name? `ModelRun` doesn't feel right, because I think one use case is to make a copy and run multiple configs or scenarios (e.g. in parallel with another instance running on another core)

znicholls:
I think smaller PRs makes sense, easier to keep track.

(You may want to retract that invitation Robert haha) I would have two
classes. One that's 'ModelRun' or 'MAGICCWorld' or something which is an
individual run and one called 'ExperimentDriver' or something which allows
you to run multiple scenarios with different configs and get output in a
specific format (which could then be passed to a plotting class to make
plots or an analysis class to do analysis or an output class to save the
output in different formats).

rgieseke:
I think I could be quite happy with this, maybe (for me) just naming the "thing" `Magicc` and then have something like

```python
magicc7 = Magicc(path="tmp/magicc", version=7)
```
or maybe two classes `Magicc6` and `Magicc7` ...

znicholls:
You can have both right? I'd probably offer an API where you can set
magicc_version=6 or 7 and other configs all at a top level but then I'd
probably implement it with a number of classes eg a
Driver/MAGICC/MAGICCWorld (which handles configs and pulling everything
together, id probably then have a MAGICC6Driver and a MAGICC7Driver as
subclasses), a Writer (which writes files, you could make SCENWriter,
CFGWriter, INEMISWriter, INCONCWriter etc. subclasses), a Reader etc. etc

The simple API would just look like a simple API but it'd call these
classes somewhere to do everything. With sensible tests we should be able
to build up the complexity without destroying the simple APIs we'll start
with.

lewisjared:
@rgieseke I renamed the `ModelRun` class to `Package` as that is a better abstraction. I also added `initialise` and `clean` methods which are not automatically run. 

We can then create Magicc6 and Magicc7 classes which inherit Package and provide some version dependent configuration, but that is outside the scope of this PR. Package is (mostly) version independent currently until we include functionality to write configuration

This has enabled:
```
package = Package()
try:
    package.initialise()
    for change in changes:
        # Write config changes
        package.run()
finally:
    package.clean()
```

This API can be improved by being able to use this package in a `with` statement which automatically initilises and cleans the package i.e.
```
with Package() as p:
    p.run()
```

@znicholls Lets not go crazy with the number of abstractions!

rgieseke:
That looks great! I broke this by merging the RCP6 renaming, so could you maybe move the class based abstraction into a separate PR -- and how about using `Magicc` or `MAGICC` as class name, that reads nicer to me _and_ we can keep most of the API when we can hook directly into MAGICC at some point without having to copy around files.

```python
magicc = Magicc()
try:
   magicc.initialise()
    for change in changes:
        # Write config changes
        magicc.run()
finally:
    magicc.clean()
```

```python
with MAGICC() as m:
    m.run()
```

In Pyhector the respective functions were called `init` and `close`:
https://github.com/openclimatedata/pyhector/blob/master/pyhector/__init__.py#L108

This comes from the C++ Hector core, is there a common Python pattern for this?

Stackoverflow discussion on PEP8 and abbreviations in class names:
https://stackoverflow.com/questions/2853531/how-do-you-pep-8-name-a-class-whose-name-is-an-acronym

I think both look good …

```python
magicc = MAGICC()
try:
   magicc.initialise()
    for change in changes:
        # Write config changes
        magicc.run()
finally:
    magicc.clean()
```

I really like were this is going!

lewisjared:
It was easier to rebase the PR than separate into 2 PRs as the compat/class depend on each other

I also refactored Package to MAGICC and moved to init/close to be consistent with pyhector

rgieseke:
@lewisjared Currently we have executable name, a filename and line number parameter, right? Will there be more differences? If not, I'd suggest getting rid of the the compat module. It should be possible to handle this in the `MAGICC` class, or as discussed in #26 in `MAGICC6` or `MAGICC7`, shouldn't it?

lewisjared:
Can we merge as is and the make that change in a separate PR. This version
is backwards compatible with 1.1 and then changes I make on Monday can be
with respect to this commit.

rgieseke:
Done -- I really like where this is heading!

znicholls:
Let's not go crazy with the abstractions yet. I think you underestimate how
many different tricks were going to need to get the control we want
(especially for file writing). If we want to stick with a principle of
'each method does one thing and each class does one job' (which I really
prefer), we're going to need a lot of classes eventually.

PR #31, opened 2018-06-03
Run flake8 via travis

lewisjared:
Automatically runs flake8 on every PR

The one rule from pep8 that I wouldn't mind breaking is line length. I typically use something like 120 chars rather than 79.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_install:
 install:
   - pip install -e .
 script:
-  - flake8 pymagicc
+  - flake8 --exit-zero
   - pytest --cov
 after_success:
   - codecov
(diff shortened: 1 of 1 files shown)


rgieseke:
Will error when there is Flake8 problem? This might be too strong ...

lewisjared:
Yes, it would return a nonzero error code. I added the --exit-zero parameter so it always returns zero and continues with the tests. But this would require someone checking the output of travis.

You can also install the pre-commit hook to spit out errors to the command line when you commit. `flake8 --install-hook=git`

znicholls:
Seems like a good idea! I'm happy with whatever settings work

rgieseke:
As it would require manual checking I don't think it's useful as is atm. We might revisit this later, for now I've added a `flake8` task to the Makefile, to encourage more frequent checking :-)
There are some checks in `flake8` that I would rather see handled on a case by case basis, not breaking CI.

In general we could consider adopting [black](https://github.com/ambv/black) to ensure a unified code style, I'd be ok with the 80 chars + 10% linewidth.

PR #32, opened 2018-06-04
Config module

lewisjared:
As a developer I want a mechanism for providing users to provide configuration parameters and a mechanism to override them.

All parameter names are case insensitive.

Default values for the configuration parameters are provided in `default_config` in `pymagicc.config`. These are overridden by user specific values in a YAML file located in the user's profile. For Windows (assuming my account name was Jared) is "C:/Users/Jared/.pymagicc.yml". Under windows this would be` ~/.config/pymagicc.yml`. Note that this file is not automatically created and is ignored if it does not exist. Finally env variables override both the default and user configurations. ENV variables are prefixed with `MAGICC_` to avoid collisions. For the `executable` parameter the environment variable specifying this would be `MAGICC_EXECUTABLE`. 

Configuration parameters can be accessed using:
```
from pymagicc.config import config
magicc_executable = config['executable']
```


rgieseke:
Hi @lewisjared 

Skip-slow looks very useful :-) Could you turn this into a separate PR?

> As a developer I want a mechanism for providing users to provide configuration parameters and a mechanism to override them.

Could you elaborate on the use-case? Would you like to essentially replace Namelists? As mentioned in the discussion on `MAGICC_EXECUTABLE` I'm very reluctant to add additional layers of config, since MAGICC already suffers from having way to many places and layers where config can be stored.

Users could read configs from namelist files, Excel, YAML or something else without Pymagicc having to explicitly support it. 
My main concern is that it becomes really hard to debug where a value is coming from.
A minor concern is that if we should be careful when mixing Pymagicc specific config (like executable path) with MAGICC's -- which could also be confusing for developers.

lewisjared:
The configuration is specific to the pymagicc package and does not add any extra configuration to MAGICC. I would expect that no one other than magicc contributors and maybe a handful of power users will need to deal with this configuration.

The main use case I see is for a user to track a version of MAGICC which is not bundled into pymagicc. They would install MAGICC somewhere and then, create a file named C:\Users\MyUserName\.pymagicc.yml. This file should then contain:
```
---
  executable: C:\Path\To\Executable
```

Then they don't have to deal with changing environment variables and pymagicc will use their new executable (when you are trying to run MAGICC6) instead of the bundled one

rgieseke:
Hmm, I'm not sure I understand the advantage -- why couldn't they do this in their Python script?

pymagicc/__init__.py
```diff
@@ -10,28 +10,25 @@
...
 import os
-import platform
 import subprocess
 
 import f90nml
 import pandas as pd
 
 from ._version import get_versions
+from .config import config as _config
+from .run import MAGICC6, MAGICC7  # noqa
```
znicholls: what does `noqa` mean?

pymagicc/config.py
```diff
@@ -0,0 +1,102 @@
+"""
+Module for collating configuration variables from various sources
+
+The order of preference is:
+Environment variable > User configuration file > Defaults
```
znicholls: does this mean it users environment variables over everything else?

pymagicc/__init__.py
rgieseke: No "quality assurance" -- tell e.g. `flake8` not to worry about this line

lewisjared: It tells the linter to ignore this line. flake8 doesn't see any code which uses `MAGICC7` so it raises an error. MAGICC7 was imported for use by package users `from pymagicc import MAGICC7`

pymagicc/config.py
```diff
@@ -0,0 +1,102 @@
...
+def parse_config_file(fname):
+    """
+    Read in a configuration file from a given filename
+    :param fname:
+    :return: The configuration in the file as a Dict, else None if the file
+        could not be found
+    """
+    if exists(fname):
+        data = yaml.load(open(fname))
+        # Convert all keys to uppercase
```
znicholls: is this just a convenience to avoid dealing with case-sensitivity?

pymagicc/config.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    data = parse_config_file(fname)
+
+    def perform_lookup(item):
+        # No file was found
+        if data is None:
+            return None
+        return data.get(item.upper())
+
+    return perform_lookup
```
znicholls: ???

znicholls: is there a missing item input argument in line 64 or do I not understand something about python?

pymagicc/config.py
lewisjared: That's the normal order of precedence. The methods of setting the parameters are in order of increasing specificity and lifespan. The default parameters in code are meant to never be changed by anyone other than the module authors. The config in the user profile is static values for a given user, finally, environment variables can be set for individual shell sessions.

pymagicc/config.py
lewisjared: It was for testing. It enables this whole function to be replaced with the expected loaded data. Otherwise, you have to mock open to return a stringified version of the configuration and it gets complicated fast!  See line 47

znicholls:
Makes sense thanks

pymagicc/config.py
znicholls: Makes sense thx

pymagicc/config.py
znicholls: ah ok I see

pymagicc/config.py
lewisjared: It is a function wrapper: https://www.thecodeship.com/patterns/guide-to-python-function-decorators/

This allows the config file to be loaded once (L91) when lookup_file is called. lookup_file returns a function which expects the same parameters as the other load functions. That function has access to the data local variable from the scope of the lookup_file function. Think of it as a function with some state.

znicholls: Very nice

znicholls:
Looks awesome to me, will add a few more comments when I get to the office in 45mins or so

lewisjared:
That path is something quite static in the most common use case of pymagicc. 

Adding it to every script that you use would be a pain. You end up having to touch several files whenever it changes. This way you get automatic upgrades across all your script for free and it is less error prone.

rgieseke:
As you said this is likely only relevant for MAGICC developers, but some of them might be doing both development work with testing *and* climate policy scenario analysis on the same machine. They might not want the MAGICC version to be changed globally without doing explicitly.

znicholls:
It's only for one session/shell instance right, not all sessions?

znicholls:
I'm not sure we'd ever want automatic upgrades across all our scripts for free, that seems to go against all ideas of permanence, reproducibility and environment isolation that i would want as a magicc user.

lewisjared:
This will allow for more fine-grained control of the configuration and provides flexibility. It would be trivial to add the functionality to look for configuration in the current folder a la Travis, Appveyor etc.

The latest commit adds the functionality to override the configuration in scripts 
```
from pymagicc import config, MAGICC6
config['executable'] = '/my/dir/executable'
assert MAGICC6.original_dir() == '/my/dir'
```

znicholls:
I am nervous because I can easily imagine myself:

1. setting up an analysis where some of the defaults are actually important
1. adding in this config file
1. having a couple of months pass
1. going back to the analysis file and having it not work and taking ages to remember this config file

I already make this mistake all the time with MAGICC's current set up and don't want to replicate it in `pymagicc` ideally.

lewisjared:
I can back out the file configuration (Just need to remove lookup_file from ConfigStore.config_lookups). This leaves explicitly specifying the config parameter, env variables and defaults.

If more configuration parameters eventuate, we can revisit user/project configuration files

lewisjared:
I do think it would be a useful feature, but we can sideline it to get this merged.

Adding back that functionality requires reverting 12bbb3d

znicholls:
The config file or having the config file in `.config`. I would like to see
the config file back in, just not assuming that path by default (the user
should have to specify the path to look in)

rgieseke:
I still don't fully see why we need to support reading from YAML in Pymagicc. There are small details to observe when using Fortran namelists or when using Python dictionaries. YAML would bring its own potential pitfalls ...
There is always the possibility to have separate modules, say `pymagicc-yaml-config`, `pymagicc-excel-config` if people wanted this.

Having a module which makes it more transparent where config comes from is certainly useful!

znicholls:
Hmm it seems to me like we want to support the following:

- default configs which just make the thing work
- the ability to change those configs
    - temporarily
    - on a more permanent basis, with a saved config file

The natural solutions to this seem to be:

- shipping a set of default configs with `pymagicc`
- allowing `pymagicc` to use environment variables (which is temporary
unless you change your `.bash_profile` or whatever in which case you should
know what you're doing enough to get yourself out of any mess you create)
- providing a way for `pymagicc` to read a more human readable config. On
this one I'm on the fence, part of me feels like python dictionaries are in
general fairly readable, part of me feels like yaml is an extra level of
easy so we should support it...

znicholls:
Hmm it seems to me like we want to support the following:

- default configs which just make the thing work
- the ability to change those configs
    - temporarily
    - on a more permanent basis, with a saved config file

The natural solutions to this seem to be:

- shipping a set of default configs with `pymagicc`
- allowing `pymagicc` to use environment variables (which is temporary
unless you change your `.bash_profile` or whatever in which case you should
know what you're doing enough to get yourself out of any mess you create)
- providing a way for `pymagicc` to read a more human readable config. On
this one I'm on the fence, part of me feels like python dictionaries are in
general fairly readable, part of me feels like yaml is an extra level of
easy so we should support it... as I've said previously, I would support putting
all our config readers, input file readers, output file readers etc. in separate
classes/modules so no complaints from me there

lewisjared:
Just to make sure we are all on the same page, the intention of the config module is to enable customisation of pymagicc. No parameters will be written to namelists for ingestion by magicc. There will only be a handful of managed parameters.

I think we all agree that pymagicc should be opinionated about the default magicc configurations.

rgieseke:
Ok, let's take a step back and understand the use case
> As a developer I want a mechanism for providing users to provide configuration parameters and a mechanism to override them.

Which parameters do we want to change? The executable and the path to a full MAGICC "package". What else?

An example usage: A MAGICC developer wants to work on a feature on the Fortran side, they make a copy (local `git clone`) and use Pymagicc to check outputs and ensure that nothing breaks while they hack on the Fortran code. While doing this Pymagicc should use the copy when running a custom testing or plotting script.

znicholls:
I think @rgieseke makes the right point and I think we should hold off on introducing complexity (although I know that word means different things to different people). What exactly are we trying to do?

Jared are you easily able to write either tests or an example of where this would be useful? I'm thinking something like:

- developer wants to be able to change MAGICC executable from the command line, this is currently not possible (I think?)
    - solution: allow the environment variable `MAGICC_EXECUTABLE` to modify `pymagicc`'s configuration
- `pymagicc` should run out of the box
    - solution: provide a set of defaults for parameters which run
- user wants to override default values
    - solution: allow the configurations to be set from within a Python script (do we also need the ability to set them from a config file or is this complexity not yet worth it)

Maybe if we take that approach things will be clearer. I'm also going to have a crack at making a template for bug fix/enhancement issues now

lewisjared:
All 3 of those use cases currently exist in this implementation.  This can wait until #38 has been merged as this will require some rework and some end to end tests demonstrating those use cases.

For the last point, that functionality was added in a later commit. Namely, you can use:
```
from pymagicc.config import config
config['EXECUTABLE'] = '/my/magicc'
magicc = MAGICC6()
assert magicc.original_dir == '/my
```

rgieseke:
I'm still not sure I fully understand the use-case/needs here ...

rgieseke:
Maybe we can re-start/clarify the discussion with more concrete user stories.

One use case I'd see would be testing a MAGICC6 binary compiled for Linux or MacOS to replace the shipped `magicc.exe`.

lewisjared:
I have rebased this again. The discussion of configuration use cases is something that can continue, but this PR is more of a tidy up then adding additional functionality. The PR has been stripped down to removing the `compat` and `path` modules in favour of a single `config` module which is easier to reason about. The `ConfigStore` class only reads environment variables and default values. Configuration can be overridden at runtime using: 
```
from pymagicc.config import config
config['EXECUTABLE'] = '/tmp/magicc'
```
More tests have been added to test the configuration functionality.

One feature that has been added is the creation of concrete `MAGICC6` and `MAGICC7` classes. In future, embedding a MAGICC7 distribution will be as simple as adding the files and a default config parameter in config.py:L16.

pymagicc/api.py
```diff
@@ -1,18 +1,16 @@
...
 
 import f90nml
 import pandas as pd
 
-from .compat import get_param
-from .paths import _magiccbinary, _magiccpath
+from .config import config
 
-_WINDOWS = platform.system() == "Windows"
+IS_WINDOWS = config['is_windows']
```
rgieseke: Why is "Windows" a config option?

pymagicc/__init__.py
```diff
@@ -301,8 +296,9 @@ def run(scenario, return_config=False, **kwargs):
         if 'endyear' in kwargs:
             year_cfg['endyear'] = kwargs.pop('endyear')
             magicc.set_years(**year_cfg)
-
-        kwargs.setdefault(get_param('emission_scenario_key'), "SCENARIO.SCEN")
+        emis_key = "file_emissionscenario" if magicc.version == 6 \
```
rgieseke: Could this check be done in the class, if run with `MAGICC6` there should be no need to check for the version again, should it? Or it could be `MAGICC` and then depend on the MAGICC version set -- but probably `run` should always use one version by default (if Pymagicc shipped with 6 and 7)

pymagicc/__init__.py
```diff
@@ -10,28 +10,26 @@
...
+from .config import config as _config
+from .api import MAGICC6, MAGICC7  # noqa
+
 __version__ = get_versions()["version"]
 del get_versions
 
-_magiccpath, _magiccbinary = _get_magicc_paths()
-
-_WINDOWS = platform.system() == "Windows"
+# default parameters and cannot be changed after module load
```
rgieseke: Can you elaborate the use case here?

Why not write a script like

```python
_magiccpath = "/tmp/magicc-super-fast"
out = run(testscen)

_magiccpath = "/tmp/magicc-test"
test_out = run(testscen)
```

pymagicc/api.py
rgieseke: We could call it "IS_WINDOWS" everywhere, btw :-)

pymagicc/__init__.py
lewisjared: True in this case the version check is not needed at all since it will always be using MAGICC6.

BTW: The MAGICC base class shouldn't be used - it is an abstract class since it doesn't define `get_executable`.

pymagicc/__init__.py
lewisjared: The less changing of module-level variables the better. Module level variables like _magiccpath don't make good configuration sources when other variables depend on them. They work great as constants, but you can end up with all kinds of dangling references to previous values. For example, changing _magiccpath doesn't change `rcp26` and the other module-level imports since they are eagerly loaded when the module is first imported.
 
In this case the configuration about the target distribution is determined when an instance of MAGICC is instantiated. Then, any variables when depending on the configuration are loaded.

My preference would be to remove _magiccpath and _magiccbinary as they no longer make sense when dealing with multiple distributions.

rgieseke: > They work great as constants, but you can end up with all kinds of dangling references to previous values. 
Very good points ... @znicholls and I discussed a bit reading IN files yesterday, and when loading these lazily (since there are many in MAGICC7), it would be good to be able to have them follow path settings as well.

If you see a way of removing _magiccpath and _magiccbinary I'd be happy with that -- also better to have fewer places to change config!

pymagicc/api.py
lewisjared: A check for windows was being performed in several different modules so it was moved to config. In my eyes, knowing what OS is being used is still configuration even if it won't change during the lifetime of the program

rgieseke:
>BTW: The MAGICC base class shouldn't be used - it is an abstract class since it doesn't define get_executable.

Is there a pythonic default to make this clearer? Maybe just `_MAGICC`?

rgieseke:
@znicholls Can you give this another look, to me this looks good to go!

lewisjared:
Or MAGICCBase

Then the api module can define
```
MAGICC = MAGICC6
```
to set the default version of MAGICC that `pymagicc` uses!

rgieseke:
Nice! Once Magicc7 becomes Public, we could easily switch.

znicholls:
Loooks good to me

PR #34, opened 2018-06-04
Make tests which run magicc as slow

lewisjared:
These tests can be skipped by passing --no-slow to pytest. The default
is to run the slow tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,17 @@
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--no-slow", action="store_true", default=False,
+        help="skip any slow tests"
+    )
+
+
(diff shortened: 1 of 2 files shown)


lewisjared:
Cherry-picked from #32

tests/conftest.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--no-slow", action="store_true", default=False,
```
rgieseke: Pandas and Xarray call this `--skip-slow` so maybe we should call it this as well.

https://pandas.pydata.org/pandas-docs/stable/install.html#running-the-test-suite

rgieseke:
Thanks a lot! I spend way too much time looking at s.l.o.w.l.y moving dots :-)

PR #35, opened 2018-06-04
Add issue templates

znicholls:


rgieseke:
This will probably need some iteration but it's a great start. Many thanks!

PR #38, opened 2018-06-06
Review API

rgieseke:
This is work-in-progress, feel free to chime in @znicholls @lewisjared


rgieseke:
This adds a couple of changes, @znicholls and @lewisjared if you could give this a look that would be great! I hope I didn't break any of the things you need, if I did please let me know!

I tried to simplify things, make names more self-explaining, tried to follow the folder structure of MAGICC7, and removed the `output_dir` functionality from `pymagicc.run` -- this is now possible with new API.

tests/test_api.py
```diff
@@ -88,27 +88,21 @@ def test_with():
...
-    p.init()
-    assert p.is_initialised()
-
-    # test a duplicated object
-    p2 = MAGICC(p.root_dir)
-    assert p2.is_initialised()
+def test_create_copy_only_once():
+    with pytest.raises(FileExistsError):
+        m = MAGICC()
+        # Create copy.
```
znicholls: I would put the mock back in here or at least have another unit test which uses the mocking framework

pymagicc/__init__.py
```diff
@@ -295,13 +289,13 @@ def run(scenario, output_dir=None,
         ``return_config`` is set to True
     """
 
-    model_run = MAGICC(output_dir)
+    magicc = MAGICC()
     try:
```
znicholls: if I'm not mistaken I think this should also work if you do something like

```
with MAGICC() as magicc:
    ...
    results = magicc.run()
```

and then you can drop the `finally` block as the `__exit__` method of `MAGICC` should take care of it

rgieseke: Good point, pushed a change!

pymagicc/api.py
```diff
@@ -39,38 +39,31 @@ def __init__(self, root_dir=None):
...
         else:
             # Create a temp directory
             self.is_temp = True
             self.root_dir = mkdtemp(prefix="pymagicc-")
 
     def __enter__(self):
-        if not self.is_initialised():
-            self.init()
+        if self.is_temp and not exists(self.run_dir):
+            self.create_copy()
```
lewisjared: Perhaps there should be a check for when a root_dir is provided but has not already been initialised.

You could either:
* Raise an exception
* Do nothing
* Create call `create_copy` (this is the previous behaviour)

pymagicc/api.py
```diff
@@ -39,38 +39,31 @@ def __init__(self, root_dir=None):
...
         return self
 
     def __exit__(self, *args, **kwargs):
-        self.close()
+        if self.is_temp:
```
lewisjared: `remove_temp_copy` already checks `is_temp == True` before deletion. Perhaps L127 can be removed in preference to checking `is_temp` in `__exit__`. This will allow a user to explicitly remove a copied directory by calling `remove_temp_copy`, but temp directories are automatically cleaned up when using the context manager.

pymagicc/api.py
```diff
@@ -89,7 +82,7 @@ def run(self, only=None):
         """
         command = [join(self.run_dir, _magiccbinary)]
 
-        if not _WINDOWS and _magiccbinary.endswith(".exe"):
+        if not _WINDOWS and _magiccbinary.endswith(".exe"):  # pragma: no cover
```
lewisjared: Is this because this statement is always true when testing under Travis?

lewisjared:
I'm happy with the proposed name changes. Can confirm that the tests still run successfully with a MAGICC7 binary so no detrimental impacts

pymagicc/api.py
rgieseke: I added it to not count it a uncovered when checking Code Coverage under a non-WIndows system.

pymagicc/api.py
rgieseke: You mean for the case when a `root_dir` is defined in

magicc = MAGICC(root_dir="/tmp/magicctest")

but there is actually no MAGICC binary and files in there?

rgieseke: That could be useful addition, as I removed the possibility to define a location for a MAGICC package in `pymagicc.run`

pymagicc/api.py
rgieseke: Testing for `is_temp` twice I didn't like as well – I thought that it should be very difficult to accidentally remove a MAGICC package, so `pymagicc` should only remove temp copies.
Maybe the first check could go?

pymagicc/api.py
lewisjared: Since this is the __enter__ method, more specifically:
```
with MAGICC(root_dir="/tmp/magicctest") as magicc:
```

The `is_initialised` check was to conform with option 3.

pymagicc/api.py
lewisjared: That functionality would better align with the method name: `remove_temp_copy`

pymagicc/api.py
rgieseke: Ok -- then
```python
def __enter__(self):
        if not exists(self.run_dir):
            self.create_copy()
        return self
```
should do it, right?

pymagicc/api.py
rgieseke: Ok, then I'll remove the first check to make this clearer.

pymagicc/api.py
rgieseke: Hm, still unsure what do here, @lewisjared 

The "promise" of using "with" is sort of to always clean up afterwards.
So now, if a user wanted to create a copy out of the MAGICC copy within Pymagicc they would have to
```python
m = MAGICC(root_dir="/tmp/magicctesting")
m.create_copy()
```
This would create a copy and even create "magicctesting" if it doesn't exist.

Is there a need to have the same behaviour with "with" (and keep the copy afterwards)?

znicholls: It seems to me that if people want a copy, they should have to make it outside of the context manager. As Robert says, if you use `with` the promise is that things are cleaned up.

How about:

- if you use `with` alongside `root_dir=`, the `magicc_dir` is cleaned up anyway (i.e. deleted) and a warning is thrown which tells the user that this is happening
- the only way to create a copy is as below

```python
m = MAGICC(root_dir="/tmp/magicctesting")
m.create_copy()
```

this would create the copy as well as the `root_dir` if it doesn't already exist

lewisjared: There isn't any promises other than the ones that are documented!
The cases outlined by @znicholls are the most logical. You would reasonably expect that the MAGICC instance exists for the life of the with block and an alternative API is provided if you want a long-lived copy. 

Perhaps we should start adding some logger messages to the library to clarify the actions being taken

rgieseke: >if you use with alongside root_dir=, the root_dir is cleaned up anyway (i.e. deleted) and a warning is thrown which tells the user that this is happening

If you had a created a copy of MAGICC, then Pymagicc shouldn't delete it I think.

Maybe 
> Raise an exception
if there is an empty `root_dir`given is the safest option.

We could expose `create_copy` as a general helper, something like:

```python
pymagicc.create_magicc_copy("/tmp/mymagicc-test")
with MAGICC(root_dir="/tmp/mymagicc-test") as m:
    m.run()

# /tmp/mymagicc-test is still around
```

rgieseke:
I think this is good enough to go, merging this now ... we can improve on it in separate PRs, especially the behaviour of
```python
with MAGICC(root_dir="/tmp/mymagicc-test") as m:
    m.run()
```

Issue #39, opened 2018-06-07
Config writing


rgieseke:
The config file writing currently done in `pymagicc.run` should become part of the API

lewisjared:
I propose adding:
* `MAGICC.write_config(filename='MAGTUNE_SIMPLE.CFG', top_level_key='nml_allcfgs', **kwargs)`
Which writes all additional arguments under the 'top_level_key' to a fortran namelist specified by the filename parameter.
e.g. `magicc.write_config(file_emisscen='RCP85.SCEN')` would write out a MAGTUNE_SIMPLE.CFG which contains the equiv of {'NML_ALLCFGS': {FILE_EMISSCEN: 'RCP85.SCEN'}}.

And a helper to create the NMLYEARS file
* `MAGICC.write_years(start_year=1765, end_year=2100, **kwargs)`

It would great to add calls for both of these functions to `MAGICC.create_copy` so a sane MAGCFG_NMLYEARS.CFG and MAGTUNE_SIMPLE.CFG is created. It would be nice if MAGICC ran out of the box with:
```
with MAGICC() as magicc:
     magicc.run()
```

rgieseke:
That looks nice! Maybe `set_config` and `set_years`, anticipating a future MAGICC version with direct API access?

PR #40, opened 2018-06-08
Config writing

lewisjared:
See #39


lewisjared:
It might be worth waiting for #38 to be merged as there will be merge conflicts

pymagicc/__init__.py
```diff
@@ -303,28 +303,16 @@ def run(scenario, output_dir=None,
...
-        all_cfgs["rundate"] = _get_date_time_string()
-
-        # Write simple config file.
-        outpath = os.path.join(model_run.run_dir, "MAGTUNE_SIMPLE.CFG")
-        f90nml.write({"nml_allcfgs": all_cfgs}, outpath, force=True)
-        # Write years config.
-        outpath_years = os.path.join(model_run.run_dir, "MAGCFG_NMLYEARS.CFG")
-        f90nml.write({"nml_years": years}, outpath_years, force=True)
+        year_cfg = {}
+        if 'startyear' in kwargs:
```
lewisjared: It would be nice to change this to the more pythonic `start_year`, but that is a slight api change

pymagicc/__init__.py
```diff
@@ -303,28 +303,16 @@ def run(scenario, output_dir=None,
...
+            year_cfg['start_year'] = kwargs.pop('startyear')
+        if 'endyear' in kwargs:
```
lewisjared: Likewise 'endyear' > end_year

pymagicc/__init__.py
rgieseke: I'd agree, but I think we should stick with the MAGICC names …

rgieseke:
Looks good to me! Thanks a lot!

Issue #41, opened 2018-06-12
Discussion: API to add read and write for HIST EMIS and CONC input files


znicholls:
**Is your feature request related to a problem? Please describe.**

At the moment you can't read or write `HIST_XXX_EMIS.IN` files or `HIST_XXX_CONC.IN` files from `pymagicc`. This is annoying because you can't see the input to `MAGICC` easily nor modify it to see what effect it has.

I think this issue should largely focus on a discussion of how we keep the metadata with the data, given that `pandas` does not yet have a good way of doing this. 

**Describe the solution you'd like**

An api which offers the ability to write and read these files. After conversations with @lewisjared and @rgieseke it's still not clear what the best way to deal with the metadata is. In my ideal world the api would be something like 

```python
MData = MAGICCInput()
MData.read('HIST_EMIS_CO2I.IN')
```

- reads in the data
- fails if it can't find the file
- fails if the file isn't in the expected format
- warning if the file has some metadata inconsistencies i.e. was written misleadingly/in an unexpected way (this one can be for stage 2 but I'm writing it down so I don't forget)

```python
MData.print_data()
```

This would print a dataframe in wide format by default, same as human readable data files, however the data would actually be stored in long format internally.

```python
MData.df
```

This would be the  long format data frame

I'm not sure what the best way to access data is given that our data is effectively multi-dimensional, perhaps

http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.xs.html
```
MData.df.xs(('CO2I_EMIS', 'WORLD'), level=['Variable', 'Region'])
```

https://pandas.pydata.org/pandas-docs/stable/advanced.html#basic-indexing-on-axis-with-multiindex
From this link: 'Note that df.loc['bar', 'two'] would also work in this example, but this shorthand notation can lead to ambiguity in general'

```python
MData.df.loc[('CO2I_EMIS', 'WORLD'),] # 
```

Custom options

```python
MData['Gas']['Region'] 
MData.df['Gas']['Region']
MData.filter(gas='Gas', region='Region') # copying pyam
```

I'm not sure what this should return, maybe a Pandas DataSeries with year as the index? We could also include units in the index..

```python
MData['Gas']
```

I think something like this should return a dataframe with all of the available regions (subject to same access question as above)

Re metadata, perhaps our class should have a metadata attribute which is a dictionary

```python
MAGICCInput.metadata
```

- prints dictionary of metadata
- warns you where metadata is missing

```python
[...]

PR #42, opened 2018-06-19
Reading Hist/IN files

lewisjared:
An implementation of the reading of a cross-version reader for MAGICC6/7 input files

#41

### DIFF
--- a/pymagicc/input.py
+++ b/pymagicc/input.py
@@ -1,0 +1,227 @@
+from os.path import basename, exists, join
+
+import f90nml
+import pandas as pd
+from six import StringIO
+
+
+class InputReader(object):
+    def __init__(self, fname, lines):
+        self.fname = fname
+        self.lines = lines
+
+    def read(self):
+        nml_end, nml_start = self._find_nml()
+
+        notes = "".join(self.lines[:nml_start])
+        metadata = self.process_metadata(self.lines[nml_start:nml_end + 1])
+
+        # Create a stream from the remaining lines, ignoring any blank lines
+        stream = StringIO()
+        cleaned_lines = [l.strip()
+                         for l in self.lines[nml_end + 1:] if l.strip()]
+        stream.write("\n".join(cleaned_lines))
+        stream.seek(0)
+
+        df, units = self.process_data(stream, metadata)
+        metadata['units'] = units
+        return metadata, df, notes
+
+    def _find_nml(self):
+        """
+        Find the start and end of the embedded namelist
+        :return: start, end indexs for the namelist
+        """
+        nml_start = None
+        nml_end = None
+        for i in range(len(self.lines)):
+            if self.lines[i].strip().startswith('&'):
+                nml_start = i
+
+            if self.lines[i].strip().startswith('/'):
+                nml_end = i
+        assert nml_start is not None and nml_end is not None, \
+            'Could not find namelist within {}'.format(self.fname)
+        return nml_end, nml_start
+
+    def process_metadata(self, lines):
+        # TODO: replace with f90nml.reads when released
+        parser = f90nml.Parser()
+        nml = parser._readstream(lines, {})
+        metadata = {
+            k.split('_')[1]: nml['THISFILE_SPECIFICATIONS'][k]
+            for k in nml['THISFILE_SPECIFICATIONS']
+        }
+
+        return metadata
+
+    def process_data(self, stream, metadata):
+        """
+        Extract the tabulated data from a subset of the input file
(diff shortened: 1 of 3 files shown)


pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+        }
+        return df, units
+
+
+class MAGICC7Reader(InputReader):
+    def process_data(self, stream, metadata):
+        gases = stream.readline().split()
+        assert gases[0] == 'GAS'
+        gases = gases[1:]
+        stream.readline()  # Unused line
```
znicholls: Not actually unused but rather only used for in certain cases. It's confusing and I'm not sure what the best thing to do with it is right now so maybe just change the comment to `TODO line, not sure how to carry this around yet so just ignoring it at the moment` and throw a warning if any value is anything other than `SET`? (I need you to explain to me how best to add commits to make these changes rather than just asking you to do them cause I'm lost at the moment).

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+        units = stream.readline().split()[1:]
+        regions = stream.readline().split()[1:]
+        index = pd.MultiIndex.from_arrays([gases, regions], names=['GAS', 'REGION'])
```
znicholls: I think index should have units too, will do an example soon

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+        return df, self._extract_units(gases, units)
+
+    def _extract_units(self, gases, units):
+        combos = set(zip(gases, units))
+        result = {}
+        for v, u in combos:
+            if v not in result:
+                result[v] = u
+            else:
+                # this isn't expected to happen, but should check anyway
```
znicholls: I love this

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+    'MAGICC6': MAGICC6Reader,
+    'MAGICC7': MAGICC7Reader,
+}
+
+
+def get_reader(fname):
+    with open(fname) as f:
+        lines = f.readlines()
+
+    # Infer the file type from the header
```
znicholls: we're going to have to be smarter about this but that's next step once we have a failing test

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class MAGICCInput(object):
+    """
+    An interface to read (and in future write) the input files used by MAGICC.
+
+    MAGICCInput can read input files from both MAGICC6 and MAGICC7. These
+    include files with extensions .IN and .SCEN7.
+
+    The MAGICCInput, once the target input file has been loaded it can be
```
znicholls: edit to become

'The MAGICCInput, once the target input file has been loaded, can be'

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+
+    >>> with MAGICC6() as magicc:
+    >>>     mdata = MAGICCInput('HISTRCP_CO2I_EMIS.IN')
+    >>>     mdata.read(magicc.run_dir)
+    >>>     mdata.plot()
+    """
+
+    def __init__(self, file_name=None):
+        """
+        Initialise a Input file object.
```
znicholls: 'an Input'

pymagicc/input.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        Allow for indexing like a pandas DataFrame
+
+        >>> inpt = MAGICCInput('HISTRCP_CO2_CONC.IN')
+        >>> inpt.read('./')
+        >>> assert (inpt['CO2']['GLOBAL'] == inpt.df['CO2']['GLOBAL']).all()
+        """
+        if not self.is_loaded:
+            raise ValueError('File has not been read from disk yet')
+        return self.df[item]
```
znicholls: super nice, this could be where we could keep units as a column but not need to specify it when getting it?

rgieseke:
That looks great!

Some minor comments:
- maybe we could unify the spelling of file_name, fname to `filename`, `filepath`?
- could the default for `read` be to use the current MAGICC instance
- maybe add an info for f90nml.reads which version will have it, e.g. >v1.0.2 (`reads` is already on master, right?)

>  (I need you to explain to me how best to add commits to make these changes rather than just asking you to do them cause I'm lost at the moment).

I tried to make edits myself, but I guess you have to check something like "allow edits"

lewisjared:
Latest commit should resolve the style and minor functionality changes. I haven't implemented the units yet

This branch is on my fork rather than openclimatedata/pymagicc so you can't make changes.

PR #44, opened 2018-06-22
Explicitly copy run directory

### LINKED ISSUE #43: Slow copying of large MAGICC distributions
**Is your feature request related to a problem? Please describe.**

Some development copies of MAGICC end up being very large due to large untracked files such as temporary build output. Using pymagicc with these directories can take several minutes to copy the large number of files which limits pymagicc utility.

**Describe the solution you'd like**

Only copy the top-level files in the run directory and create the out directory rather than copying it. The configuration in the run directory is currently flat and anything not in the root run directory are build artifacts.

**Describe alternatives you've considered**

* Whitelisting files to copy - this seems like it could cause headaches to maintain a list of white lists which may differ between MAGICC6 and MAGICC7
* Blacklisting - That causes too much coupling between the development processes of magicc and pymagicc

lewisjared:
Copies all files (excluding subdirectories) in run and bin folders. Bin folder currently doesn't exist in any distribution, but could in the near future. All files in the root of the MAGICC distribution are also copied

Fixes #43

### DIFF
--- a/pymagicc/api.py
+++ b/pymagicc/api.py
@@ -1,8 +1,7 @@
 import shutil
 import subprocess
-from distutils import dir_util
 from os import listdir, makedirs
-from os.path import basename, dirname, exists, join
+from os.path import basename, dirname, exists, join, isfile
 from tempfile import mkdtemp
 
 import f90nml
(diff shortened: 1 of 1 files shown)


rgieseke:
Great, thanks for taking the time to time the results!

PR #45, opened 2018-06-22
Don't automatically create a temp directory on object instantiation

lewisjared:
Delay the creation of the temporary directory until `create_copy` is
called. `run_dir` and `out_dir` now return None, until `create_copy` is
called.

### DIFF
--- a/pymagicc/api.py
+++ b/pymagicc/api.py
@@ -43,10 +43,9 @@ class MAGICC(object):
         else:
             # Create a temp directory
             self.is_temp = True
-            self.root_dir = mkdtemp(prefix="pymagicc-")
 
     def __enter__(self):
-        if self.is_temp and not exists(self.run_dir):
+        if self.is_temp and self.run_dir is None:
             self.create_copy()
         return self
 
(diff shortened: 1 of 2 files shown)


rgieseke:
With the longer MAGICC7 copy times I see that this could be a problem, but what is the use-case for having a `MAGICC` instance without a copy?

I could imagine that if IN-files were exposed one could do something like

```python
magicc = MAGICC()
magicc.histrcp_co2_conc
```

But you might have something else in mind?

lewisjared:
That temporary directory won't be automatically cleaned up which is a bug more than anything. Nothing is copied, rather a temporary folder is created everytime an object is instantiated which is unintended.

Also, you can get the current default magicc directory with `MAGICC6().original_dir`

rgieseke:
Alright, good catch and thanks for the explanation! 

With 
```python
with MAGICC6() as magicc:
    pass
```
you get everything automagically, with
```python
magicc = MAGICC6()
magicc.create_copy()
```

you do everything yourself.

I had forgotten about that ...

PR #46, opened 2018-06-22
Search for tags in the header of input files

lewisjared:
Add additional metadata from the header
Related to #41


rgieseke:
There is some metadata in MAGICC6 as well, isn't it?

E.g. `HISTRCP_CO2B_EMIS.IN`

```
DATA:  Historical landuse CO2 (CO2B) Emissions (HISTRCP_CO2B_EMIS) 
SOURCE:  RCP data as presented on http://www.iiasa.ac.at/web-apps/tnt/RcpDb, August 2009 
CONTACT:  Base year emissions inventories: Steve Smith (ssmith@pnl.gov) and Jean-Francois Lamarque (Jean-Francois.Lamarque@noaa.gov); RCP 3-PD (IMAGE): Detlef van Vuuren (detlef.vanvuuren@pbl.nl); RCP 4.5 (MiniCAM): Allison Thomson (Allison.Thomson@pnl.gov); RCP 6.0 (AIM): Toshihiko Masui (masui@nies.go.jp); RCP 8.5 (MESSAGE): Keywan Riahi (riahi@iiasa.ac.at); Concentrations & Forcing compilation: Malte Meinshausen (malte.meinshausen@pik-potsdam.de)
```

rgieseke:
This has gotten a bit messy :-) ... you could undo the commits on your branch and force push to clean it up again. Or we close this PR and start a new one with header data only.

In general I don't mind the occasional additional cleanup commit in a PR but we should try to keep one feature to one PR.

Contentwise, you get Pandas and f90nml when you do `pip install -e .` For a while Pymagicc was pinned to a certain f90nml version that's why I kept this out of the "dev-requirements" -- so you can install a specific Pymagicc version into the dev virtualenv or run the latest master version.

A check for wine could be very useful in the Makefile but it should be a separate task (you can play with some of Pymagicc without wine). Would sure be nice to make it simpler for someone to go from `git clone` to `make test` and have a local dev setup!

tests/test_input.py
```diff
@@ -104,11 +104,15 @@ def test_header_metadata():
...
+    assert m6.process_header('Data: Average emissions per year') == {'data': 'Average emissions per year'}
+    assert m6.process_header('DATA:  Historical landuse BC (BCB) Emissions (HISTRCP_BCB_EMIS) ') == {'data': 'Historical landuse BC (BCB) Emissions (HISTRCP_BCB_EMIS)'}
+    assert m6.process_header('CONTACT:   RCP 3-PD (IMAGE): Detlef van Vuuren (detlef.vanvuuren@pbl.nl); RCP 4.5 (MiniCAM): Allison Thomson (Allison.Thomson@pnl.gov); RCP 6.0 (AIM): Toshihiko Masui (masui@nies.go.jp); RCP 8.5 (MESSAGE): Keywan Riahi (riahi@iiasa.ac.at); Base year emissions inventories: Steve Smith (ssmith@pnl.gov) and Jean-Francois Lamarque (Jean-Francois.Lamarque@noaa.gov) ') == {'contact': 'RCP 3-PD (IMAGE): Detlef van Vuuren (detlef.vanvuuren@pbl.nl); RCP 4.5 (MiniCAM): Allison Thomson (Allison.Thomson@pnl.gov); RCP 6.0 (AIM): Toshihiko Masui (masui@nies.go.jp); RCP 8.5 (MESSAGE): Keywan Riahi (riahi@iiasa.ac.at); Base year emissions inventories: Steve Smith (ssmith@pnl.gov) and Jean-Francois Lamarque (Jean-Francois.Lamarque@noaa.gov)'}
+    # assert warning on this one
+    # m6.process_header('DATE: 26/11/2009 11:29:06; MAGICC-VERSION: 6.3.09, 25 November 2009')
```
lewisjared: Why is this raising a warning?

lewisjared:
This is ready for review. I don't think raising warnings for missing metadata will be useful as that is simcap/magicc's problem not pymagicc.

znicholls:
My thought was that we have many files which are written without proper metadata/don't conform to the ideal format and it would be good to know when that is for two reasons:

1. you won't get all the metadata you would expect back
1. you know which files need to be checked

Hence it's not really a simcap/magicc/anything problem, it's a problem for the user who expects things to be a certain way. Happy to be overruled on this one though...

znicholls:
I would prefer not to put it in 1.2 as I thin we want to change the way units are handled and don't want to confuse users

lewisjared:
The change to the structure of the units should be done in #49

lewisjared:
@rgieseke Instead of rebasing, the conflict is easy enough to resolve in merge

rgieseke:
@znicholls Can you merge this?

znicholls:
I think I've done the right thing...

tests/test_input.py
znicholls: The idea here was that if someone tried to read a file that had the wrong metadata format, they would get told and so would know that if they tried to access e.g. metadata['date'], it wouldn't be there

rgieseke:
Thank you!

PR #49, opened 2018-06-22
Change unit read in and dataframe handling

znicholls:
This will change the units from being read in as metadata to being read in with the dataframe. This helps make the most of pandas' automatic index alignment which seems wise as discussed in #42


znicholls:
@lewisjared @rgieseke I'm going to put aside a couple of hours later in the week to copy over my readers and writers from my matlab scripts if that's ok for you? I think that will work better than going along the current trajectory as the input files are pretty messy and I think the solution I (eventually) worked out was a good one

znicholls:
This is ready to go (for the most part, have to fiddle with final few things). Have a look at the example notebook, it's not using `pint` yet but can once we have `pint` with units. 

It would be good to merge this, then I can rebranch to add SCEN file handling stuff.

rgieseke:
Alright, this is quite some changes, I'll have a look soon!

.travis.yml
```diff
@@ -9,7 +9,7 @@ before_install:
   - sudo dpkg --add-architecture i386
   - sudo apt-get update
   - sudo apt-get install --no-install-recommends -y wine
-  - pip install pytest pytest-cov codecov mock nbval matplotlib
+  - pip install pytest pytest-cov codecov mock nbval matplotlib git+https://github.com/thedataincubator/expectexception.git@master
```
znicholls: `expectexception` means that we can put cells which raise errors in our example notebooks but still have them under sensible CI

pymagicc/definitions/__init__.py
```diff
@@ -0,0 +1,17 @@
+from os.path import dirname, join
+
```
znicholls: This module will hold all of our MAGICC definitions. It's useful for writing files as well as for unit conversions later on.

pymagicc/input.py
```diff
@@ -8,6 +8,11 @@
 from six import StringIO
 
 from pymagicc import MAGICC6
+from .definitions import (
+    dattype_regionmode_regions,
+    emissions_units,
+    concentrations_units,
+)
```
znicholls: I've basically completely rewritten our `input` module. The idea is that we have a topline object, `MAGICCInput`, which holds our dataframes. It has the metadata attribute as well as a simpler direct access thing than having to always remember how to do pandas indexing. 

Then we have `InputReader`, our superclass for reading files (this could actually be moved into its own file in future). We then subclass `InputReader` to make readers for specific files as necessary e.g. `HIST_CONC_INReader` reads historical concentration files (which are all, in theory, the same format). 

Same idea with writers for `InputWriter` and subclasses. 

The functions `_get_writer` and `_get_reader` choose which class to use to read a file, based off the filename (I think this is the only way to do this, introspecting more heavily is likely a nightmare given internal contradictions in files... happy to change though).

Idea is to extend to reading and writing more file types in future.

tests/test_input.py
```diff
@@ -11,13 +11,7 @@
 import f90nml
```
znicholls: Tests are probably the easiest thing to read and the most sensible place to start (alongside the example notebook!)

rgieseke:
Ok, bikeshedding question first:

What about `HistConcInReader` vs. `HIST_CONC_INReader`?

Should we try to stick to Python conventions or make it easier for people who have memorised the MAGICC conventions.
I'd be in favor of sticking to Python rules ....

znicholls:
Python rules. I'm just lazy but will gladly make those changes. I'm
guessing all the class names should go to camelcase?

It's odd that black isn't changing it?

rgieseke:
Maybe https://github.com/PyCQA/pep8-naming can be used to check. 

With Acronyms it's probably hard to do well with `black`.

znicholls:
Ok. On these matters I yield to whatever you think is best.

znicholls:
Good question, will fix that..

znicholls:
It could also be a good change. I wasn't sure what the best way was. I had
in mind that most people would probably look at the csvs in excel and so
having a yuck csv format was ok. Changing to storing as python variables
would also be fine, but I think we'd have to provide the interface to look
at the data in a table for easy comparison.

rgieseke:
> It could also be a good change. I wasn't sure what the best way was. I had
in mind that most people would probably look at the csvs in excel and so
having a yuck csv format was ok. Changing to storing as python variables
would also be fine, but I think we'd have to provide the interface to look
at the data in a table for easy comparison.

The project could include both -- CSV for Excel and other projects, auto-generated Python Dicts or whatever makes sense in the module. But we can really keep this for later.

znicholls:
- [ ] fix header line split
- [ ]  new issue for csv storage format
- [ ] fix lines in emissions_unit
- [ ] change numeric tests to use allclose
- [ ] fix class names to match PEP8

znicholls:
> Why is the header field in split lines with "\n"?

sorry looked at this properly now. That's because 'header' is simply all of the stuff before the namelist without any processing. It's a way of making sure we keep all the text that was in the file without having to try and process every line we don't recognise.

znicholls:
@rgieseke this should cover it off, up to you to merge

rgieseke:
Alright, being bold ... and merging before the last green light!

znicholls:
> Alright, being bold ... and merging before the last green light!

have you lost your mind?! What are they teaching you over there...

rgieseke:
> have you lost your mind?! What are they teaching you over there...

Probabilistic CI, first to greenlight wins!

PR #50, opened 2018-06-23
Update Makefile with test target

znicholls:
@rgieseke can you check this and maybe add to CI please?


rgieseke:
Added a fix for fish users :-)

Nice!

PR #54, opened 2018-06-27
Rename MAGICC baseclass to MAGICCBase and expose MAGICC6 as MAGICC

lewisjared:
I won't die in a ditch whether or not to expose a MAGICC class (which defaults to MAGICC6) as the default implementation of the API.

Either way, MAGICCBase is a better name as it is an abstract base class and cannot be directly instantiated.

### DIFF
--- a/pymagicc/__init__.py
+++ b/pymagicc/__init__.py
@@ -26,6 +26,9 @@ from .api import MAGICC6, MAGICC7  # noqa
 __version__ = get_versions()["version"]
 del get_versions
 
+# Default to using MAGICC6 for now
+MAGICC = MAGICC6
+
 # default parameters and cannot be changed after module load
 _magiccpath, _magiccbinary = MAGICC6().original_dir, MAGICC6().original_dir
 
--- a/pymagicc/api.py
+++ b/pymagicc/api.py
@@ -27,7 +27,7 @@ def _copy_files(source, target):
             shutil.copy(full_filename, target)
 
 
-class MAGICC(object):
+class MAGICCBase(object):
     """
     Provides access to the MAGICC binary and configuration.
 
@@ -213,14 +213,14 @@ class MAGICC(object):
                                stepsperyear=12)
 
 
-class MAGICC6(MAGICC):
+class MAGICC6(MAGICCBase):
     version = 6
 
     def get_executable(cls):
         return config['executable']
 
 
-class MAGICC7(MAGICC):
+class MAGICC7(MAGICCBase):
     version = 7
 
     def get_executable(cls):


rgieseke:
Agreed on MAGICCBase!
So how strong is "die in a ditch" ...? ;-)
I think it might be good to always just be explicit re MAGICC6/7 ...

lewisjared:
Meaning that I didn't mind either way. I would agree about being explicit, but it a nice to have since we have the same interface for both.

rgieseke:
Just worried about people using it in a science context and suddenly their scripts where they didn't pin dependencies get different results when they update Pymagicc at some point in the future ...

@znicholls thoughts?

znicholls:
Are we trying to make a backwards compatible update or not?

I hate the idea of making MAGICC = MAGICC6. I would much rather give MAGICC
the ability to work out whether it is running 6 or 7 based on the
executable or force people to always ve explicit about their version.

rgieseke:
I'm not worried about backwords compatibility ... we 3 can probably still figure it out ... 

My only concern is really people re-running a script with a global Pymagicc install that suddenly gives slightly different results and they can't figure out why ...

In the future we would probably it to MAGICC = MAGICC7 so I'll remove the mapping for now.

znicholls:
If they can't figure out how to manage their environments I feel like there's nothing we can do about that... If we bump to version 2.0 (major increment to indicate backwards incompatible change) then what more can we do?

I'm still nervous about defaulting to MAGICC7, I feel like if you're going to use MAGICC, you should at least be aware that there are different versions and hence have to chose one?

rgieseke:
Sure, now you have to be explicit

```python
with MAGICC6() as magicc6:
    # ...
with MAGICC7() as magicc7:
    #
```

How we communicate a change in `run` we can talk about once we have a publicly available MAGICC7 :-)
We can consider having a separate function `run7` if that's safer ... depends also on how much results differ for standard scenarios ...

znicholls:
Sorry I only just realised how this issue started. I also like the idea of exposing `MAGICCBase` as it shows that the interfaces are the same (at least I think this would be the point?)

Will we continue to maintain a standalone `run`? Or will you only be able to run magicc with a `MAGICC6` or `MAGICC7` instance? 

Results differ enough to warrant being careful...

rgieseke:
I like to keep the `run` functionality, I use it a lot.

znicholls:
Ok could we just add a keyword argument to `run` then? Something like `MAGICC_EXECUTABLE` or `MAGICC_VERSION` which defaults to `MAGICC_6`?

PR #55, opened 2018-06-27
Exception on missing executable

lewisjared:
The current failure mode if MAGICC6/MAGICC7 is misconfigured is very opaque. It raises an exception when a call to `os.path.exist` gets a `None`. 

This adds an explicit check when `create_copy` is called and raises a sane exception. 2 lines of code change 80 lines of tests!

### DIFF
--- a/pymagicc/api.py
+++ b/pymagicc/api.py
@@ -80,11 +80,16 @@ class MAGICCBase(object):
     def __exit__(self, *args, **kwargs):
         self.remove_temp_copy()
 
+    def get_executable(self):
+        raise NotImplementedError
+
     def create_copy(self):
         """
         Initialises a temporary directory structure and copy of MAGICC
         configuration files and binary.
         """
+        if self.executable is None or not isfile(self.executable):
+            raise FileNotFoundError('Could not find MAGICC{} executable: {}'.format(self.version, self.executable))
         if self.is_temp:
             assert (
                 self.root_dir is None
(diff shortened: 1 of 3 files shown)


pymagicc/api.py
```diff
@@ -66,11 +66,16 @@ def __enter__(self):
...
+    def get_executable(self):
+        raise NotImplementedError
+
     def create_copy(self):
         """
         Initialises a temporary directory structure and copy of MAGICC
         configuration files and binary.
         """
+        if self.executable is None or not isfile(self.executable):
+            raise ValueError('Could not find MAGICC{} executable: {}'.format(self.version, self.executable))
```
rgieseke: How about FileNotFoundError

https://docs.python.org/3/library/exceptions.html#FileNotFoundError

rgieseke: Now that we're Python3 only we can use Python3 exceptions :-)

tests/test_config.py
```diff
@@ -47,11 +73,11 @@ def test_simple_config():
     assert config['EXECUTABLE'] == default_config['EXECUTABLE']
 
 
-def test_precendence(env_var):
+def test_precendence(env_override):
```
rgieseke: Should be "precedence"?

znicholls: yes, precendence is not a word

znicholls:
Nice to have in v1.2, probably not essential..

znicholls:
@lewisjared thoughts on best way to merge? rebase?

lewisjared:
Conflicts are rather minor and only affect the tests. I'll resolve the conflicts in a manual merge tonight so I can check that the tests still pass

lewisjared:
I actually think that I have an unpushed commit resolving the two comments above so I can append to resolve the conflicts

znicholls:
@lewisjared @rgieseke can we merge this?

lewisjared:
@znicholls Rebased and good to go. Sorry about the delay

rgieseke:
Many thanks!

PR #68, opened 2018-07-03
Raise FileNotFoundError when specified root_dir is not found

### LINKED ISSUE #62: Test for initialising with root_dir
We haven't tested this case and it's not clear what happens

lewisjared:
Fixes #62


rgieseke:
Thanks!

PR #70, opened 2018-07-08
Diagnose TCR and ECS

znicholls:
The input api is still a bit nasty but this functionality is useful so I figure we can bring this in and I can fix the api in #49


rgieseke:
This looks super nice! I'd maybe replace "Your TCR ECS results" with "The TCR ECS results" (as in "not your fault" ...

Maybe it's also easier to always return the full result set and not have the added complexity of `full_results`.

znicholls:
Where is 'Your TCR ECS results'? 

> Maybe it's also easier to always return the full result set and not have the added complexity of full_results.

I thought about that but I figured that mucking around working out which year is the 70th year of increase etc. was simple functionality I could provide so just did it...

rgieseke:
> Where is 'Your TCR ECS results'?

In the potentially very useful error messages, should have written [results], sorry.

> I thought about that but I figured that mucking around working out which year is the 70th year of increase etc. was simple functionality I could provide so just did it...

Maybe we can have both, useful summary and full results by default?

znicholls:
Looks good now

PR #72, opened 2018-07-09
Rename MAGTUNE_SIMPLE to MAGTUNE_PYMAGICC

znicholls:
Fairly simple. It's obviously a bit annoying that you have to change the name in 3 places (`set_config`, `MAGCFG_USER.CFG` and `default_config.nml`) but I don't think we can avoid such issues easily. A fix requires more thought than I've given it so I'm not going to suggest anything now. 

I seem to have also removed whitespace from a bunch of lines, maybe it's time for an `.editorconfig`?


rgieseke:
This seems useful, but is now really the only file with settings from Pymagicc?

znicholls:
I think that it, for now, is the only file. Unless you also want to rename `SCENARIO.SCEN` to `PYMAGICC_SCENARIO.SCEN`?

lewisjared:
The trailing spaces are only cleaned up once the file has been touched so over time the inconsistencies should correct themselves.

rgieseke:
We just discussed this a bit: should `set_config` actually allow for configurable filenames (since the main file hardcodes the value ...

In the MAGICC test setup it's used in `generate_plots.py`

```python
magicc.set_config("MAGTUNE_XTRAPARAS_THISRUN.CFG", **config)
```

So maybe it should remain, @znicholls ?

znicholls:
Ah yes very sensible oops. I guess we just move that comment about which filename to use from the actual writing method into the run method?

rgieseke:
Not sure I understand, moving to the run function? Why is the xtra_paras needed in the tests (or anyway ...)

znicholls:
As usual with magicc config this is complicated, next week haha

PR #77, opened 2018-07-11
Put notebooks under CI

znicholls:
This seems to be one way to do it... I'm hoping @lewisjared might have some ideas about how to cache conda envs so CI time doesn't explode

@rgieseke you may wince at the extra installation overhead this creates so I'm happy to also have this be vetoed


rgieseke:
Hmm, I didn't check fully yet, but maybe something simpler like https://blog.thedataincubator.com/2016/06/testing-jupyter-notebooks/ could also work?

znicholls:
Under the hood it's the same thing.

However yes you probably could do seomething clever by pip installing
jupyter and hence skip conda?

znicholls:
Or maybe not I don't know..

My other thought was that if we use scons we could potentially show chains
of notebooks in future...

.travis.yml
```diff
@@ -9,10 +9,22 @@ before_install:
...
+  # thanks https://github.com/ghego/travis_anaconda_jupyter/blob/master/.travis.yml
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  - conda info -a
+
+  - conda env create -n test-env-pymagicc python=$TRAVIS_PYTHON_VERSION -f environment.yaml
```
lewisjared: You could have a play with caching the conda/pip caches and miniconda downloads which may shave some time off the run

.travis.yml
```diff
@@ -9,10 +9,22 @@ before_install:
...
+  - source activate test-env-pymagicc
   - pip install pytest pytest-cov codecov mock
 install:
   - pip install -e .
 script:
   - pytest --cov
+  - scons
```
lewisjared: Perhaps this only needs to be run on py3.6 if test duration is a problem.

rgieseke:
I fully agree that an automatic check would be super nice but i'm very reluctant to get a bigger machinery ...

rgieseke:
Closing this for now in favor of #83

PR #78, opened 2018-07-11
Make diagnose tcr ecs api more useful

znicholls:
Short term fix to make the diagnosis notebook slightly nicer @rgieseke


znicholls:
When #79 has a solution, these changes would need to be tweaked again

PR #79, opened 2018-07-11
Keeping track of config state

znicholls:
The persistance test I've put at the end fails when it shouldn't. The solution requires `MAGICCBaseClass` to somehow 'remember' its config so that when the `diagnose_tcr_ecs` method is called it doesn't overwrite the previous settings.

I'm not sure what the best solution is here. I feel like we could either:

- add a `_config` property to `MAGICCBaseClass`
    - pros: in memory hence fast
    - cons: more places for config to hide

- force the `set_config` routine to read in the current `MAGTUNE_SIMPLE.CFG` everytime it's called
    - pros: config in only one place
    - cons: reading and writing files is slow

Then I'm not sure what the behaviour should be:
- if you're setting a previously unset variable, same as current
- if you're setting a previously set variable should we have:
    - same as current (i.e. silent overwrite)
    - warning about overwriting (which could be silenced?)
    - something else?

Decided to just Jared's new API which does the job for us (setting previously set variable silently overwrites which I'm ok with as querying config is easy).

# Show that keeping track of config state works

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


znicholls:
Once #136 goes through, we can use the `update_config` method and then start to hold config in memory, only writing to disk as part of `run`

PR #80, opened 2018-07-12
Autoformat in black

rgieseke:
Adds an autoformat task and unifies coding style using the **Black** formatter.

@znicholls @lewisjared Any open PRs should be able to be mergeable after a rebase and `make black`.


rgieseke:
Ok, maybe changing won't be that simple, but in principle, running `black` manually should enable clean merges later.

PR #83, opened 2018-07-12
Test notebooks with `nbval`

rgieseke:


rgieseke:
This seems simpler than #77 -- happy to revise with inclusion of scons etc. if there is a need for nested notebooks @znicholls

PR #94, opened 2018-08-28
Few things I noticed

znicholls:
Very brief, just adding tcr/ecs diagnosis tests into master

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -201,8 +201,8 @@ def test_diagnose_tcr_ecs_config_setup(mock_set_years, mock_set_config, magicc_b
     mock_set_years.assert_called_with(startyear=1750, endyear=4200)
     mock_set_config.assert_called_with(
         FILE_CO2_CONC="TCRECS_CO2_CONC.IN",
-        RF_TOTAL_RUNMODUS="CO2",
         RF_TOTAL_CONSTANTAFTERYR=2000,
+        RF_TOTAL_RUNMODUS="CO2",
     )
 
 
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -389,3 +389,198 @@ def test_missing_config(config_override):
...
+@pytest.mark.slow
+def test_integration_diagnose_tcr_ecs(package):
+    actual_result = package.diagnose_tcr_ecs()
+    assert isinstance(actual_result, dict)
+    assert "tcr" in actual_result
+    assert "ecs" in actual_result
+    assert actual_result["tcr"] < actual_result["ecs"]
+    if isinstance(package, MAGICC6):
+        assert (
+            actual_result["tcr"] == 1.9733976000000002
```
rgieseke: That's probably too accurate ... :-)
Maybe there should be an "almost_equal" comparison or so be used?

znicholls: for a stable MAGICC version (i.e. when this is always done with pymagicc's version of MAGICC6), that should never change right? Hence the accuracy is ok/a useful backup to make sure we didn't swap the binary somewhere?

znicholls:
I find the mocking useful because it means I can write my functions bit by bit. For example, it means I can write a test of e.g. `diagnose_ecs_tcr` that just tests whether `diagnose_ecs_tcr` calls e.g. `setup_diagnosis`, `run_magicc`, `process_results` without actually having to have written the details of those functions yet. I find that useful as I can write my tests as I go, rather than needing to have everything written before any test will pass. It's also useful as it tells me where things are failing exactly e.g. without the mock the test of `diagnose_ecs_tcr` would fail if MAGICC didn't run (because diagnosing ecs and tcr relies on calling MAGICC at some point). However with the mock, it won't fail if MAGICC doesn't run (because MAGICC is never called in the tests thanks to the mock).

If you want to know more about mocking https://medium.com/@bfortuner/python-unit-testing-with-pytest-and-mock-197499c4623c

znicholls:
> looks good to me otherwise

Merge at will

tests/test_api.py
rgieseke: Not sure, I think the "...0002" is just coming from Python reading the data and this might be annoying to debug in CI if it changed randomly.

znicholls: smart

rgieseke:
Awesome, thanks!

Issue #95, opened 2018-08-29
Add tests that nml in files is correct


znicholls:
**Describe the bug**

At the moment we don't have any way to check that e.g. `THISFILE_FIRSTDATAROW` is set correctly. I'd like to write a script which does this and then just run this over the entire `run` directory to see what happens. 

**Failing Test**

No code here, conceptual at the moment. 

**Expected behavior**

Reading in a data file which has an incorrectly set namelist should at least raise a warning.

PR #102, opened 2018-08-31
Add read and write for SCEN file in new style

znicholls:
@rgieseke quite a lot going on here, are you happy if we use these new style readers and writers once they're properly tested, deleting your old ones?

Do you want to keep the old interfaces or get rid of them completely/provide 'new format' scenarios?


znicholls:
#95 is then the next step to add the tests that these writers are actually working rather than just being self-consistent with the pymagicc API

rgieseke:
What's with

```python
class OpticalThicknessInWriter(InputWriter):
    pass
```

rgieseke:
Why do we use `codecs.open` in `input.py`?

Maybe using the default `open` with `newline=""` could fix the failing Windows tests.

```
newline controls how universal newlines works (it only applies to text mode). It can be None, '', '\n', '\r', and '\r\n'. It works as follows:

    On input, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newline mode is enabled, but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.
    On output, if newline is None, any '\n' characters written are translated to the system default line separator, os.linesep. If newline is '', no translation takes place. If newline is any of the other legal values, any '\n' characters written are translated to the given string.
```

https://docs.python.org/release/3.2/library/functions.html#open

rgieseke:
I'll try this ...

rgieseke:
Ok, using `newlines` doesn't help. I guess we need to get the writing right as well ...

Maybe we should use a string comparison function which ignores newlines?
Thoughts @znicholls ?

znicholls:
> What's with
> 
>> class OpticalThicknessInWriter(InputWriter):
>>     pass

this makes `OpticalThicknessInWriter` the same as `InputWriter` (strictly speaking a subclass but they'll behave identically). If I don't have the `pass` I get a `SyntaxError` and I like setting it up this way because it makes it obvious where I put my methods if `OpticalThicknessInWriter` needs to change from the default behaviour of `InputWriter` in some way.

> Why do we use `codecs.open` in input.py?

So that umlauts in author names in headers are read correctly (it was the only way I could see to do it...)

It's odd that `newline` doesn't work...

rgieseke:
> It's odd that newline doesn't work...

Another test on-going ...

rgieseke:
One Windows build green ... seems the Umlauts also work with `open` and the proper encoding set ...

rgieseke:
Not sure, if enforcing `\n` always is a bad idea for Windows MAGICC users ...

Issue #30, opened 2018-06-01
Automatic linting


rgieseke:
- automatic linting (PEP8 and docstyle) would be useful
- either automatic on PRs or at least documented and callable from the `Makefile`

znicholls:
Callable from the `Makefile` is done thanks to you, do you want to add it on PRs or shall we close?

rgieseke:
Flake8 needs to be configured to respect `black` defaults

rgieseke:
Maybe like this: https://github.com/ambv/black/blob/master/.flake8

znicholls:
> Flake8 needs to be configured to respect black defaults

I don't understand, where does Flake8 come into it?

rgieseke:
Yes, automatic might be an option but i think it shouldn't block merging

znicholls:
> Yes, automatic might be an option but i think it shouldn't block merging

so what do we do with this issue then?

znicholls:
Given we now have black and flake8 linting, do we close this or block merges that fail? I think we should chose one or the other.

rgieseke:
Let's close this!

PR #103, opened 2018-08-31
Fix flake8 warnings

rgieseke:
See #30


rgieseke:
@znicholls Last commit could probably also be done as a one-liner in the next line ... your choice.

`make flake8` doesn't report problems any more.
Can you check and merge?

rgieseke:
They have overlapping but different purposes. Flake8 also checks for stylistic issues, black just formats blindly.

Issue #117, opened 2018-09-07
Docs to track


znicholls:
Which docs should we track in the repo?

rgieseke:
Can you elaborate?

znicholls:
there's like a zillion things in the `docs` and `docsrc` folders, do we want them all in the repo?

rgieseke:
`docsrc`:

api.md     pymagicc.api.md     pymagicc.input.md
index.md@  pymagicc.config.md  pymagicc.md

The md files are auto-generated so one could argue they don't need to exist in the repo ... Not sure if they help with debugging output problems. 
+1 for one for removing and ignoring them

znicholls:
ok can you add whatever we should ignore/remove in #118, I still don't really understand what we should/shouldn't keep

rgieseke:
`docs`:

this should contain everything that is needed to display the site, both for being served via GitHub pages and locally for people who cloned the repo.

If anything is not used or needed it can go.

rgieseke:
Will do!

znicholls:
yep closed by #118

Issue #114, opened 2018-09-07
Add documentation on how to build docs


rgieseke:
Continued from #108

rgieseke:
Just noting things down, something more coherent should be developed from this

Uses Pydocmd which uses MkDocs 

https://github.com/NiklasRosenstein/pydoc-markdown

Config is in https://github.com/openclimatedata/pymagicc/blob/master/mkdocs.yml

docs/index.html: pymagicc/*.py
	pydocmd simple pymagicc+  > docsrc/pymagicc.md
	pydocmd simple pymagicc.api+ > docsrc/pymagicc.api.md
	pydocmd simple pymagicc.input++ > docsrc/pymagicc.input.md
	pydocmd simple pymagicc.config > docsrc/pymagicc.config.md
	mkdocs build

Currently index.md in docsrc is a symbolic link to the Readme.
We could have additional "prose" files in this directory.
Or link stuff like CONTRIBUTING, CHANGELOG

The api needs some work as many functions don't show up yet (don't know whether due to missing docstrings or due to lacking config)

Does that help @znicholls ?

PR #118, opened 2018-09-07
Doc generation automation

znicholls:
A good enough for now start on #114 

Raises #117


znicholls:
> Does docs/sitemap.xml.gz need to be there?

I have no idea, I just adding everything that wasn't being ignored by `.gitignore`...

Happy to adjust as you want

znicholls:
have pushed what I get, still have the `gz` file but have added to `.gitignore`, happy to merge when u r

rgieseke:
Alright, I seem to have an older Mkdocs version.

rgieseke:
Ok, so using the newer version I don't get any differences any more.
It's probably fine to ignore the .gz file and i prefer not to have too many binary files in the repo :-)

PR #128, opened 2018-09-11
Add new section to docs

rgieseke:


rgieseke:
This should add your new section ... need to add a task to clear out the docs dir before re-building ...

Issue #129, opened 2018-09-11
Simplify Doc building


rgieseke:
Maybe we should have a commit-hook to re-build the docs?

Should we have the master branch reflect the latest released version? What's available at https://openclimatedata.github.io/pymagicc/ is what is on Master but isn't what people pip installed.
We could consider having a dev branch. 

Or, probably a good solution too, have building the docs a part of "publish-on-pypi" packaging ...

Locally, one can always build the docs if needed.

I'd be in favor of the "on-update" i think.
Thoughts @znicholls

znicholls:
I quite like the idea of having a `dev` branch. That way what people see when they land on `master` is what they get from `pypi` and what they get if they use the latest release and it's clear just how far things are behind (although this isn't what pandas does so might fail that test of yours...)

Then I'd just have building the docs be part of publish-on-pypi, as you say  can always build locally for dev purposes.

rgieseke:
I like the idea of a `dev` branch too, but I think rapid Pymagicc development won't continue forever, i expect it to stop as MAGICC7 stabilises ... Having a `dev` branch would likely complicate things for the occasional contributor so I'd think we should stick with `master` and have frequent releases (as we currently do) ...

rgieseke:
I'll look into displaying the version number in the API docs, this should be useful ...

rgieseke:
Closing, done in #130

Issue #116, opened 2018-09-07
Add contributing guide


znicholls:
Another one for the todo list

PR #130, opened 2018-09-12
Extend docs

rgieseke:
Pinging @znicholls 
What else should we add here, more details on development/contributing probably?
As discussed in #129 and #116.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Master
 =====
 
-- (#108)[https://github.com/openclimatedata/pymagicc/pull/108] Added ability to read all files in MAGICC6 run folder (`pymagicc/MAGICC6/run`) to a common format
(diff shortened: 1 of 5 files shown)


znicholls:
I think that's basically all of it. Once we have those you could then pick it up, look at the examples, read through contributing and developing and add something, completing the full pipeline?

rgieseke:
I've added a "Developing" section, not sure we need more than the "Contributing" currently in Readme, what do you think?

znicholls:
> not sure we need more than the "Contributing" currently in Readme, what do you think?

Also not sure we need it right now. I think the aim will be something like what we have in magicc or in scmcallib but doesn't need to be immediate

rgieseke:
Alright ... if you're happy with the current state, we can merge.

rgieseke:
Would be good if we (=you) do the next release together and add anything that's missing/not working well.

PR #132, opened 2018-09-15
Add new packages and data to setup.py

znicholls:


rgieseke:
Looks good to me, there is no new data however, right?

znicholls:
sorry I mean the csvs in definitions...

rgieseke:
Ahh, i missed this line :-)

PR #133, opened 2018-09-17
Move definitions into a reusable place

znicholls:


pymagicc/definitions/__init__.py
```diff
@@ -15,3 +15,54 @@
 # TODO: do read ins for these too
 emissions_units = {}
 concentrations_units = {}
+
+magicc6_emms_code_all_emissions = [
```
rgieseke: A comment on what these are would be helpful ... Code 0, 1, PRN are not obvious to non Magicc core devs i guess

znicholls: If I put them in a function will they get picked up by the docs so these defintions appear in the docs too (which is really where they belong)?

rgieseke: Good point, one could document them in a separate page in the docs or we switch to Sphinx or something else which can document module constants ...

PR #139, opened 2018-09-27
Out file read in

znicholls:
I'll update this later. For now, @lewisjared this branch has all my output file reading stuff. It should give some clues about which files which should be able to read (I don't think it should be all, some output files are super weird...) and give you a set of readers with which you can check that any binary reader is behaving sensibly. 

Edit: tidying up and doing this properly now
# Pull request

- [x] Tests added
- [x] Documentation added
- [x] Example added (either to an existing notebook or as a new notebook, N/A here, will sort in #154 )
- [x] Description in `CHANGELOG.rst` added
- [x]  check that old API still returns a `dict`
- [x]  new style returns new style Dataframe

## Adding to CHANGELOG.rst

```
```


lewisjared:
Just pushed a commit adding the ability to read binary output for 4Box files. Still loads of test failures, but reading BINOUT and OUT files now produce dataframe containing the same results.

I think we need to move to a 2 level MultiIndex for the output files which contains Variable and Region as indexing these dfs are not simple. We could write the units (if available) as a property to the series. This can be changed to make the series units aware once your contribution to pandas is released.

As for the TODO level, I have no idea what that is, but I think it more metadata for series rather than an index as no file don't have 2 different values for TODO for the same variable and region (I don't think at least). The same is probably true for UNITS - there aren't 2 time series for a given variable and region that have different units.

znicholls:
> Just pushed a commit adding the ability to read binary output for 4Box files. Still loads of test failures, but reading BINOUT and OUT files now produce dataframe containing the same results.

Wicked will have a look tomorrow in detail.

> I think we need to move to a 2 level MultiIndex for the output files which contains Variable and Region as indexing these dfs are not simple.

The wrapper for `MAGICCData` means you should be able to simply index with e.g. `MAGICCData[variable, region]` in whatever order and get a data frame back. See [the example](https://github.com/openclimatedata/pymagicc/blob/master/notebooks/Input-Examples.ipynb)

I really want to avoid getting rid of the other metadata. It is super useful for being able to map between SCEN7 files and other file types and will be important as we try to move towards a uniform input format. With the accessing wrapper, I think the utility of these features outweighs the cost of having that extra wrapper.

Having these lines does add a different issue for binary files, namely that they don't have any units metadata. I'm not sure how to deal with this, I'd rather not hard code the units but don't see how else to do it. I think the solution comes with tweaks to MAGICC itself (so that we can somehow query units when reading binary files/access a stored array which documents expected units for all variables), not the output file format unless you see another solution?

lewisjared:
Can we make a better MAGICC bin format for MAGICC7 which includes more metadata. The files should also contain a version identifier to differentiate between old/new and different variables.

znicholls:
I think that is an excellent idea!

pymagicc/io.py
```diff
@@ -611,6 +654,145 @@ def _read_notes(self):
...
+
+        column_headers["variables"] = [
+            "OCEAN_TEMP_" + l for l in column_headers["regions"]
+        ]
+        column_headers["regions"] = [region] * len(column_headers["regions"])
+
+        return column_headers, metadata
+
+
+class BinData(object):
```
znicholls: why is this public?

pymagicc/io.py
```diff
@@ -611,6 +654,145 @@ def _read_notes(self):
...
+
+        assert (
+            self.data[self.pos + 4 + size : self.pos + 4 + size + 4].cast("i")[0]
+            == size
+        )
+        self.pos = self.pos + 4 + size + 4
+
+        res = np.array(d.cast(t))
+
+        # Return as a variable or a numpy array if it is an array
```
znicholls: 'return as a scalar or a numpy array if it is an array'

tests/test_io.py
```diff
@@ -1039,5 +1312,55 @@ def test_can_read_all_files_in_magicc6_dir(file_to_read):
...
+    for p in invalid_out_files:
+        if re.match(p, file_to_read):
+            valid_filename = False
+    # Skip files not marked as invalid
+    if valid_filename:
+        return
+
+    mdata = MAGICCData(file_to_read)
+    try:
+        with pytest.raises(NoReaderWriterError):
```
znicholls: For these files, raise a custom error message saying we're not planning on ever being able to read them because their format is so weird

tests/test_io.py
```diff
@@ -1039,5 +1312,55 @@ def test_can_read_all_files_in_magicc6_dir(file_to_read):
...
+        mdata_bin.read(TEST_OUT_DIR)
+    except InvalidTemporalResError:
+        # Some BINOUT files are on a subannual time scale and cannot be read (yet)
+        return
+
+    mdata_ascii = MAGICCData(file_to_read.replace('BINOUT', 'OUT'))
+    mdata_ascii.read(TEST_OUT_DIR)
+
+    # There are some minor differences between in the dataframes due to availability of metadata in BINOUT files
+    mdata_ascii.df.columns = mdata_ascii.df.columns.droplevel('UNITS').droplevel('TODO')
```
znicholls: change this so that the binary files come back with units unknown and TODO equal to SET

lewisjared:
Sure. It would be great to get this merged soon. I'll add another commit resolving your comments. `pymagicc.io` should probably be split up into a package with some submodules one day as it is getting complex.

I have made a few changes to the binary readers so to handle a new (yet to be determined) format, but these commits should go in a different PR.

pymagicc/io.py
lewisjared: Everything in Python is public?? I'll refactor the name to `_BinData`.

znicholls: sorry you know what I mean, thanks

znicholls:
@rgieseke all yours

PR #143, opened 2018-10-13
Only read in PARAMETERS.OUT if it exists

lewisjared:
# Pull request

- [x] Tests added
- [x] Description in CHANGELOG.md added

Fixes #132


lewisjared:
Note that this is a breaking change.

The `config` attribute used to initialise to an empty dict, but now it is `None` until a valid parameter file is read. This feels more sane to me

PR #146, opened 2018-10-13
Docs fixups

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
- (#XX)[http://link-to-pr.com] Added feature which does something
- (#XX)[http://link-to-pr.com] Fixed bug identified in (issue)[http://link-to-issue.com]
```


rgieseke:
Looks good to me, only one question about the path fiddling:

@swillner Why was the path fiddling necessary? https://github.com/openclimatedata/pymagicc/pull/146/commits/6f42943ad82e32d1bd122452aa8b285141d857dd

If the path is not set Sphinx explicitly will use whatever Pymagicc version is installed, right?

znicholls:
I think the default Sphinx config file is out of date. Maybe that path fiddling was a good idea in the past but no longer I think. 

> If the path is not set Sphinx explicitly will use whatever Pymagicc version is installed, right?

Yep/whatever it finds when it imports pymagicc, which is the behaviour we want I think (given devs should be working in a virtual env)

rgieseke:
> Yep/whatever it finds when it imports pymagicc, which is the behaviour we want I think (given devs should be working in a virtual env)

Good point, and the version should be stated clearly in the template.

What happens on ReadTheDocs?

znicholls:
> What happens on ReadTheDocs? (Edit: Which version wil it use?)

Good question. pip installable latest?

znicholls:
Turns out it explodes (https://readthedocs.org/projects/netcdf-scm/builds/7940589/), undoing that change then

rgieseke:
Alright I'm merging this then ...

rgieseke:
@znicholls There are settings on Readthedocs (I've added you as a maintainer) for requirements files etc. -- but I think having this set in `conf.py` is probably easier to find (than a setting on RTD).

PR #154, opened 2018-10-16
Move to openscm conventions

znicholls:
# Pull request

- [x] Tests added
- [x]  check that old API still returns a `dict` (moved to #161)
- [x]  new style returns new style Dataframe (moved to #161)

## Adding to CHANGELOG.rst

```
```


rgieseke:
Column headers convention:
Lowercase, underscore, follow Pyam

Variable and acronym naming convention:
use (and enhance) ghgrenamer (https://github.com/openclimatedata/ghgrenamer), "Emissions|N2O|MAGICC AFOLU"

'B' becomes" MAGICC AFOLU"
'I' becomes "MAGICC Fossil and Industrial"

Regional split:
World|Northern Hemisphere|Land
World|R5...
(copy Pyam)

znicholls:
> Regional split:
> World|Northern Hemisphere|Land

'World' rather than 'Global' here is jarring but it's better to be consistent with Pyam than precious

znicholls:
from earlier discussions:

- move to long format only (in preparation for using pyam capability)

znicholls:
@rgieseke `Atmospheric Concentrations|CO2` or `Concentrations|CO2`, I think the argument for the former is in case we one day introduce an ocean pH module to MAGICC and need `Ocean concentrations` or similar

znicholls:
@rgieseke this one is pretty massive. Tried to keep it just to the minimum I needed, unfortunately that's quite a lot...

CHANGELOG.rst
```diff
@@ -4,6 +4,8 @@ Changelog
 master
 ------
 
+- (`#154 <https://github.com/openclimatedata/pymagicc/pull/154>`_) Change to using openscm variables for all user facing data as well as preparing to move to using OpenSM dataframes
```
rgieseke: Should be OpenSCM throughout
```suggestion
- (`#154 <https://github.com/openclimatedata/pymagicc/pull/154>`_) Change to using OpenSCM variables for all user facing data as well as preparing to move to using OpenSM dataframes
```

pymagicc/io.py
```diff
@@ -415,8 +472,54 @@ def _read_magicc6_style_header(self, stream, metadata):
 
         return column_headers, metadata
 
+    def _get_column_headers_update_metadata(self, stream, metadata):
```
rgieseke: What does this do? Get headers or update metadata?

docs/index.rst
```diff
@@ -29,6 +29,8 @@ Pymagicc
     pymagicc.api
     pymagicc.io
     pymagicc.definitions
+    pymagicc.config
+    pymagicc.utils
```
rgieseke:
```suggestion
```
Remove module with only private functions

rgieseke: It claims that you could add a commit with this directly ... can see a button to do this?

rgieseke: Or alternatively one could add something general to `pymagicc.utils.rst`

pymagicc/io.py
znicholls: both haha

PR #162, opened 2018-10-19
Add tests of compatibility

znicholls:
# Pull request

- [x] Tests added
- [ ] Documentation added (where applicable)

## Adding to CHANGELOG.rst

```
```


znicholls:
@rgieseke all yours

pymagicc/io.py
```diff
@@ -1045,13 +1045,28 @@ def find_region(x):
...
+        The MAGICC version for which we want to write files. MAGICC7 and MAGICC6
+        namelists are incompatible hence we need to know which one we're writing for.
+
+    _scen_7 : bool
+        Whether this writer writes SCEN7 files or not. We need this as SCEN7 files
+        have a unique definition of what the THISFILE_REGIONMODE flag should be set to
+        which is contradictory to all other files.
+    """
+
+    _magicc_version = 7
```
rgieseke: Why are here both a magicc and a scen file flag?

scripts/read_write_scen.py
```diff
@@ -0,0 +1,13 @@
+import os
```
rgieseke: What is this script for? To test write and read back?

If for testing, could also go in `tests`.

Some comment would be  ...

pymagicc/io.py
znicholls: do the comments not explain it?

znicholls: yes it can :'(

PR #163, opened 2018-10-19
156 hfc 245

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


rgieseke:
Seems a useful change and clarification!

znicholls:
it's harder than I thought, give me an hour

znicholls:
ok @rgieseke all yours (checks pending)

znicholls:
OpenSCM without hyphen based on our conversation the other day where you
said it would be better to avoid hyphens in variable names where possible
(although we also have spaces eg 'magicc afolu' so maybe that's a waste of
time and consistency with whatever pyam does is more important).

MAGICC7 without hyphen and all uppercase because having hyphens in fortran
is asking for trouble and because fortran is case insensitive and all
uppercase makes it more obvious.

rgieseke:
Always trade-offs :-) 
In general, I prefer simple usability as Pandas DataFrames columns (no spaces, no hyphens), but it can be useful to stay consistent with other variable names. The current one is fine though!

Issue #120, opened 2018-09-09
Discussion: lazy loading MAGICCInput


znicholls:
@lewisjared I find the lazy loading of MAGICCInput really confusing. I'm sure there is a super useful reason for it but I can't see it, can you just run me again through why it's helpful?

lewisjared:
What about it is confusing? Is it the extra function call before they can be used? 

Lazy loading allows objects to refer to a specific filename without having a concrete magicc instance defined. `MAGICCData('RCP26.SCEN7')` refers to the configuration file which contains emissions for RCP2.6 independent of the MAGICC distribution being used. This object can be created in a module before a specific MAGICC instance exists. If the files were read at the time of the initialisation of the object, you could not do this because you don't have an absolute path to the file. Also what if the file doesn't exist and you actually want to create it?

The logic for reading a file is encapsulated in the `read` routine so if we want to synchronously read the files when the object is constructed, the call just has to be added to the `__init__` routine. It is also easy enough to call the read method automatically when accessing the data if it has not been explicitly called.

znicholls:
> What about it is confusing? Is it the extra function call before they can be used?

No not quite. My thought was that `read` and `write` should have the same function signature i.e. `read(path, filename)` or `read(filename, path)`. I just couldn't quite work out how to make that logic fit with the lazy loading logic.

Do these signatures make sense to you?

```python
def read(self, filepath=None, filename=None):
    """
    Read an input file from disk

    # Parameters
    filepath (str): The directory to read the file from. This is often the
        run directory for a magicc instance. If None is passed,
        the run directory for the version of MAGICC6 included in pymagicc
        is used.
    filename (str): The filename to read. Overrides any existing values.
        If None is passed, the filename used to initialise the MAGICCData
        instance is used.
    """
```

```python
def write(self, filepath=None, filename=None):
    """
    Write an input file to disk

    # Parameters
    filepath (str): The directory to write the file to. This is often the
        run directory for a magicc instance. If None is passed,
        an error is raised.
    filename (str): The filename to write. Overrides any existing values.
        If None is passed, the filename used to initialise the MAGICCData
        instance is used.
    """
```

rgieseke:
Instead of `filepath=None` couldn't we have `filepath=pymagicc._MAGICCBINARY`?

(upper-casing the var to emphasize it's a constant)

[...]

PR #165, opened 2018-10-21
Move to one backend

### LINKED ISSUE #161: Move to one backend, not two
This will be major breaking change of v2.0.0, makes our lives simpler re maintenance and more consistent for users (also us).

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

This PR should follow #162

Closes #161 #120


znicholls:
@rgieseke I can't work out how to fix the widgets in this

rgieseke:
Sorry, GitHub is having issues again ...

pymagicc/io.py
```diff
@@ -61,9 +61,9 @@
 """
 
 
-def _unsupported_file(filename):
+def _unsupported_file(filepath):
     for uns in UNSUPPORTED_OUT_FILES:
-        if re.match(uns, filename):
+        if re.match(uns, filepath):
```
rgieseke: "uns" is too German a word, let's call it "outfile" or something ...

pymagicc/__init__.py
```diff
@@ -176,168 +70,44 @@ def read_scen_file(scen_file):
...
-    Return a timestamp with current date and time.
-    """
-    now = datetime.datetime.now()
-    return now.strftime("%Y-%m-%d %H:%M")
-
-
-def write_scen_file(
-    scenario, path_or_buf=None, description1=None, description2=None, comment=None
-):
+def run(scenario, magicc_version=6, out_parameters=True, **kwargs):
```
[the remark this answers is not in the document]
znicholls: that's probably wise, will change to `return_parameters`

pymagicc/api.py
```diff
@@ -332,19 +394,21 @@ def set_years(self, startyear=1765, endyear=2100):
...
         """Writes the output configuration
 
-        There are a number of configuration parameters which control which variables are written to
-        file and in which format. Limiting the variables that are written to file can greatly speed
-        up the running of MAGICC. By default, calling this function without specifying any variables
-        will disable all output.
+        There are a number of configuration parameters which control which variables
+        are written to file and in which format. Limiting the variables that are
+        written to file can greatly speed up the running of MAGICC. By default,
+        calling this function without specifying any variables will disable all output.
```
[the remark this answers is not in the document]
znicholls: yes but all the `output_xx` variables are set to zero so nothing will get written

pymagicc/api.py
```diff
@@ -147,27 +153,68 @@ def out_dir(self):
...
+            )
+
+        usr_cfg = read_cfg_file(join(self.run_dir, "MAGCFG_USER.CFG"))
+        top_tuning_model_key = sorted(
+            [k for k in usr_cfg["nml_allcfgs"] if k.startswith("file_tuningmodel")]
+        )[-1]
+        top_tuning_model = usr_cfg["nml_allcfgs"][top_tuning_model_key]
+        if top_tuning_model != "PYMAGICC":
+            raise ValueError(
+                "PYMAGICC is not top the tuning model in `MAGCFG_USER.CFG`, your run is likely to fail/do odd things"
```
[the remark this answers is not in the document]
znicholls: Hmm yes this is cryptically named and thinking about it more badly done. I'll change, in the meantime, some context, some of which is not new but I put it all here to try and make sure it's clear (Alex Nauels may also have helpful thoughts as the SLR config is something else again...). Having written this, I'm happy to put this comment in the MAGICC flags docs. 

When you run MAGICC, you have a series of `.CFG` files which set parameter values. The first two are always (and must be) `MAGCFG_DEFAULTALL.CFG` and `MAGCFG_USER.CFG`. Exception number 1: in the compiled binary that is shipped with Pymagicc, the file it looks for is `MAGCFG_DEFAULTALL_69.CFG`, not `MAGCFG_DEFAULTALL.CFG`.

After these two compulsory files, you can then specify extra `.CFG` files. Each of these `.CFG` files is specified in `FILE_TUNINGMODEL_X` flags.

The first confusing thing: say you have `FILE_TUNINGMODEL_X=ABCD`, MAGICC will look for `MAGTUNE_ABCD.CFG`. Note that MAGICC will not look for `ABCD` or `ABCD.CFG`. Hence if we have `FILE_TUNINGMODEL_8=PYMAGICC`, then the 8th tuning file MAGICC will look for will be `MAGTUNE_PYMAGICC.CFG`. This is why we write the configuration to `MAGTUNE_PYMAGICC.CFG` and set `FILE_TUNINGMODEL_X=PYMAGICC`.

The second confusing thing: If `FILE_TUNINGMODEL_X` is set to `USER`, MAGICC won't look for any configuration file at all and will simply just move to the next `FILE_TUNINGMODEL_X` flag. I think the rationale here is that you have to read `MAGCFG_USER.CFG` (see above) and hence using this flag as a "don't read" flag is safe. Fortunately if you set `FILE_TUNINGMODEL_X` to an empty string i.e. `""` then MAGICC will also skip that `.CFG` file so there is a safer and more obvious option.

The third confusing thing: the first flag it looks at is `FILE_TUNING_MODEL`, without any number, whilst the rest are all of the form `FILE_TUNINGMODEL_X` where X is a positive integer (i.e. not zero).

The fourth confusing thing: the convention above changes in MAGICC7, where `FILE_TUNINGMODEL` is replaced by `FILE_TUNINGMODEL_1`

The fifth confusing thing: the maximum value of `X` in `FILE_TUNINGMODEL_X` varies depending on your binary and there's no way to query the binary except for just trying higher and higher `X` until it fails.

The sixth confusing thing: each `.CFG` file contains a set of parameters, each of which overwrites any previously read parameter values. i.e. the file pointed to by `FILE_TUNINGMODEL_4` overwrites flags set by `FILE_TUNINGMODEL_3` etc. However, each `.CFG` file can contain `FILE_TUNINGMODEL_X` flags itself. Hence you could have the following situation:

1. `MAGCFG_DEFAULTALL.CFG` sets `FILE_TUNINGMODEL=USER` and `FILE_TUNINGMODEL_X=USER` for all X i.e. no overwrites on top of what will be read from `MAGCFG_DEFAULTALL.CFG` and `MAGCFG_USER.CFG`
1. `MAGCFG_USER.CFG` sets e.g. `FILE_TUNINGMODEL=ZNEXAMPLE` and `FILE_TUNINGMODEL_2=PYMAGICC`, leaving everything else untouched
    - hence at this point we would expect MAGICC to read in `MAGCFG_DEFAULTALL.CFG`, then overwrite its values with values in `MAGCFG_USER.CFG`, then overwrite those values with values in `MAGTUNE_ZNEXAMPLE.CFG`, finally overwriting those values with the values in `MAGTUNE_PYMAGICC.CFG`
1. `MAGTUNE_ZNEXAMPLE.CFG` sets `FILE_TUNINGMODEL_2=USER`
    - now `MAGTUNE_PYMAGICC.CFG` won't be read at all

The reason this is confusing/annoying is that you have to read, and trace, the hierarchy of every single `.CFG` file in order to work out what is going to happen. I've done a small function which does this in the `tests` directory of the MAGICC repo (we need it to reproduce CMIP6 concs), if someone wants then make an issue and I can put it in here.

To avoid any really unexpected, silent surprises, we want the pymagicc cfg file, `MAGTUNE_PYMAGICC.CFG` (which gets altered to `MAGTUNE_PYMAGICC.CFG` internally by MAGICC) to overwrite everything else. I'll put a solution which actually does this properly up now.

Whilst we're here, these 'conventions' gets worse when we compare to what happens with the emission scenario files. 

In MAGICC6 it's simple, there's only one emissions scenario file and it goes in `FILE_EMISSIONSCENARIO=DFGH`. MAGICC then looks for files that match `DFGH` and `DFGH.SCEN`.

In MAGICC7, there are now multiple `FILE_EMISSCEN_X` flags (note the shift from `FILE_EMISSIONSCENARIO`). The values found in the file specified in each `FILE_EMISSCEN_X` flag overwrite any previously read in values.

Confusing things here: 

The first emissions scenario file is specified by `FILE_EMISSCEN` (without the number), which matches the MAGICC6 convention for `FILE_TUNINGMODEL` but contradicts the MAGICC7 convention for `FILE_TUNINGMODEL`.

Say we have `FILE_EMISSCEN=DFGH`, MAGICC7 looks for files that match `DFGH`, then `DFGH.SCEN7` and then `DFGH.SCEN` (in that order). Hence the first emissions scenario can be `SCEN7`, or `SCEN`, with preference being given to `SCEN7` files if there are two scenario files with the same stem (i.e. `RCP26.SCEN7` is chosen before `RCP26.SCEN` if `FILE_EMISSCEN=RCP26`). 

All other `FILE_EMISSCEN_X` files can only be `SCEN7` files. The rationale here (I think) is that `SCEN` files don't contain easy to read metadata hence overwriting with them is difficult/dangerous.

If you set `FILE_EMISSCEN_X=NONE` then MAGICC will just move on to the next `FILE_EMISSCEN_X` flag. However, from above it's clear (and I've tried it) that if you set `FILE_TUNINGMODEL_X=NONE`, MAGICC will look for `MAGTUNE_NONE.CFG`, not find it and blow up. Hence there's a direct contradiction there too.

Of course the final part is that each `.CFG` file can overwrite the `FILE_EMISSCEN_X` flags of previous `.CFG` hence working out which scenario will actually be run is also not trivial.

@lewisjared this may be of interest to you

pymagicc/api.py
```diff
@@ -147,27 +153,68 @@ def out_dir(self):
...
+            )
+
+        if scenario is not None:
+            scen_name = "SCENARIO.SCEN"
+            self.write(scenario, scen_name)
+
+            # TODO: work out a way to deal with MAGICC7's nested overwrite stuff
```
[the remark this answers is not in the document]
rgieseke: Viva passed I'd say!

Just a quick thought ... we don't have to support anything that MAGICC7 does in Pymagicc if it's too ... complicated?

pymagicc/api.py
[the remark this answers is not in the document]
rgieseke: Viva passed I'd say!

Just a quick thought ... we don't have to support anything that MAGICC7 does in Pymagicc if it's too ... complicated?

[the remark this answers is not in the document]
znicholls: > Viva passed I'd say!

This is material for your PhD proposal.

> Just a quick thought ... we don't have to support anything that MAGICC7 does in Pymagicc if it's too ... complicated?

No I think what we do is set things up to run in the simplest way possible and then just fail if they're not how we want them (I'm doing a quick modification now which should make how this can work clear). Step 2 is to get all this nonsense removed from MAGICC7 before we release it (which is also possible as we can now easily move all the pre-processing into Pymagicc where debugging doesn't require you to learn how to use a Fortran debugger :D).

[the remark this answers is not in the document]
rgieseke: +1 for Step 2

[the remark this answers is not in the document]
znicholls: @matthiasmengel @anauels this may interest you.

@rgieseke just added an even  more confusing example

pymagicc/api.py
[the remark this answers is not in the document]
znicholls: will do so in #168

znicholls:
Alrighty @rgieseke let's go for round 2 of review please?

rgieseke:
Started reading ... quick question: Did you compare the content of PARAMETERS.OUT (in addition to un-changed results for the RCPs?)

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
...
+  - Breaking the API, hence requiring significantly re-writing the tests to match the new API, bumping the major version number and updating the examples.
+  - Locking up Pymagicc so that it will only run if MAGICC's ``.CFG`` files are configured in the simplest way possible (see :ref:`MAGICC flags`). This required re-writing the ``pymagicc/MAGICC6/run/MAGCFG_USER.CFG`` file that ships with Pymagicc (although the result is the same, as confirmed by the fact that the outputs of the four RCPs are unchanged in ``tests/test_pymagicc.py``).
+  - Adding a function to pull a single configuration file from a MAGICC ``PARAMETERS.OUT`` file to aid the transition to the change referred to above (i.e. one could run MAGICC with whatever config elsewhere and then get a single config file which can be used with Pymagicc from the resulting ``PARAMETERS.OUT`` file).
+  - Tidying up the docs to make linking a bit simpler and more reusable.
+  - Only passing ``filepath`` (i.e. the combination of path and name) to reading/writing functions to remove ambiguity in previous language which used ``file``, ``filepath``, ``path``, ``name`` and ``filename``, sometimes in a self-contradictory way.
```
rgieseke: Great!

docs/magicc_flags.rst
```diff
@@ -0,0 +1,215 @@
...
+- ``MHALO_ADJSTFUTREMIS2PAST_0NO1SCALE``: adjust future Montreal Protocol halogen emissions to past emissions.
+
+
+Configuration flags
+-------------------
+
+When you run MAGICC, you have a series of ``.CFG`` files which set parameter values.
+The first two are always (and must be) ``MAGCFG_DEFAULTALL.CFG`` and
+``MAGCFG_USER.CFG``. *Exception:* in the compiled binary that is shipped with
+Pymagicc, the file it looks for is ``MAGCFG_DEFAULTALL_69.CFG``, not
```
rgieseke: An aside, but does the 69 stand for Version 6.9?

docs/magicc_flags.rst
```diff
@@ -0,0 +1,215 @@
...
+        return cfg
+
+    # Example usage
+    run_dir = "/here/there/MAGICC/run"
+    namelist_to_derive = "all_cfgs"
+    output_file = "./somewhere/else/example.cfg"
+    derive_final_cfg(run_dir, namelist_to_derive).write(output_file)
+
+
+To avoid any really unexpected, silent surprises, we want the pymagicc ``.CFG`` file,
```
rgieseke: Very good :-)

docs/magicc_flags.rst
```diff
@@ -0,0 +1,215 @@
...
+``FILE_EMISSCEN_X`` files can only be SCEN7 files. The rationale here (I think) is
+that SCEN files don't contain easy to read metadata hence overwriting with them is
+difficult/dangerous.
+
+*Gotcha:* If you set ``FILE_EMISSCEN_X=NONE`` then MAGICC will just move on to the
+next ``FILE_EMISSCEN_X`` flag. However, from above it's clear that if you set
+``FILE_TUNINGMODEL_X=NONE``, MAGICC will look for ``MAGTUNE_NONE.CFG``, not find it
+and blow up. Hence there's a direct contradiction there too.
+
+*Gotcha:* Of course the final part is that each ``.CFG`` file can overwrite the
```
rgieseke: Maybe Pymagicc should only support SCEN7 and have a conversion function (if possible)?

pymagicc/api.py
```diff
@@ -147,27 +152,65 @@ def out_dir(self):
...
         Parameters
         ----------
+        scenario : :obj:`pymagicc.io.MAGICCData`
+            Scenario to run. If None MAGICC will simply run with whatever config has
+            already been set.
+
         only
             If not None, only extract variables in this list
 
+        include_parameters
```
rgieseke: Whatever "none" means what is set in the config file, right? Maybe this could be spelled out, if i understood it correctly?

pymagicc/io.py
```diff
@@ -1867,5 +1868,91 @@ def read_cfg_file(fullfilename):
...
+
+    Returns
+    -------
+    :obj:`f90nml.Namelist`
+        An f90nml object with the cleaned, read out config.
+
+    Examples
+    --------
+    >>> cfg = pull_cfg_from_parameters_out(magicc.metadata["parameters"])
+    >>> cfg.write("/somewhere/else/ANOTHERNAME.txt")
```
rgieseke: Why ".txt"? And not ".cfg"?

docs/magicc_flags.rst
znicholls: I believe so although of course it's never that clear

docs/magicc_flags.rst
znicholls: That's an option although at the moment it basically only supports taking in dataframes (and then does the conversions internally, or will do after #168) so it's kind of a non-issue?

pymagicc/api.py
znicholls: Not quite. Basically you need a default which says, don't change the existing config on my `MAGICC` instance. i.e. if you already said somewhere else:

```python
magicc = MAGICC6()
magicc.set_config(out_parameters=0)
```

then you don't want `magicc.run()` to silently set `out_parameters`, you want it to just not change any of the existing config. Hence you need the default to be something which is neither `True` nor `False` hence I went for `None`. Line 208 is where it's used (I'm not sure how better to explain this in the docs...):

```python
if include_parameters is not None:
    kwargs["out_parameters"] = 1 if include_parameters else 0
# else don't alter whatever the current config value of out_parameters is
```

znicholls: suggestions welcome

pymagicc/io.py
znicholls: dumb example, my bad

znicholls:
> Looks good to me! Didn't manage to read the tests yet ... feel free to merge!

There's no rush (about to meet with Matt) so take your time

docs/magicc_flags.rst
rgieseke: Ok, seems to be non-issue then ...

pymagicc/api.py
rgieseke: Hmm, that's still confusing to me ... 

Basically it's just syntactic sugar for haveing "out_parameters` = 1 as a keyword argument in the run function?

Then we could also drop it, and put a reminder in the doc string that, for example, out_parameters=0 will prevent reading out parameters?

znicholls: > Basically it's just syntactic sugar for haveing "out_parameters` = 1 as a keyword argument in the run function?

Yep

znicholls:
@rgieseke 3rd time lucky

.gitignore
```diff
@@ -1,3 +1,6 @@
+# User generated config files
```
rgieseke: Maybe change this to "Config files generated from testing script ``pull_cfg..." 

This script has a path from your home in it ...

rgieseke:
One thing ...
It seems that the new version is slower with the defaults by about 0.5 s:

# This branch
 %time pymagicc.run(pymagicc.rcp26)                                     
CPU times: user 1.47 s, sys: 123 ms, total: 1.6 s

# Pymagicc 1.3.2
%time _ = pymagicc.run(pymagicc.rcp26)                                 
CPU times: user 1.05 s, sys: 87.9 ms, total: 1.14 s

znicholls:
> It seems that the new version is slower with the defaults by about 0.5 s:

Yes I suspect my writer is not a very fast way of doing things. Get it working first, optimise second?

rgieseke:
Totally! Merge and let's see how fast we can get it ...

PR #177, opened 2018-11-02
Fix SCEN read in bug

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


pymagicc/io.py
```diff
@@ -618,14 +618,19 @@ def read_data_block(self):
...
-                    self._read_data_header_line(self._stream, "YEARS")
-                )
+                pos_vars = self._stream.tell()
+                variables = self._read_data_header_line(self._stream, "YEARS")
             except IndexError:  # tried to get variables from empty string
                 break
             except AssertionError:  # tried to get variables from a notes line
-                break
+                try:
+                    # just in case...
```
rgieseke: Can you elaborate on the "just in case"? :smile:

znicholls: probably should read, `# check that the header line isn't labelled with "YEAR" instead` although I think there's a better way to do this so will refactor in 20 mins

PR #187, opened 2018-12-19
Building scenarios

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


lewisjared:
rebase this as a lot is already committed to master

znicholls:
Done

CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 master
 ------
 
+- (`#187 <https://github.com/openclimatedata/pymagicc/pull/187>`_) Added `pymagicc.io.join_timeseries`
```
rgieseke: Maybe add something like "simplifies joining/merging scenarios to create custom scenarios"?

rgieseke:
The "Building Scenarios" is super useful, but do we have some documentation on the "todo" stuff already somewhere? Might be useful to elaborate or add a  link to docs where this is explained a bit more.

pymagicc/io.py
```diff
@@ -2095,3 +2119,106 @@ def get_generic_rcp_name(inname):
     except KeyError:
         error_msg = "No generic name for input: {}".format(inname)
         raise ValueError(error_msg)
+
+
+# TODO: move out of here in clean up
```
rgieseke: Create an issue, do it or somehow make it clearer how this could be re-factored?

pymagicc/io.py
```diff
@@ -2095,3 +2119,106 @@ def get_generic_rcp_name(inname):
...
+def join_timeseries(base, overwrite, join_linear=None):
```
rgieseke: `join_linear` - should this be a tuple (while the code would still accept a list) or rather, `start`, `stop` - i think we usually have start and end points as separate args (correct me if i'm wrong or if there is a specific reason to have a list)

znicholls: I think having separate start and end is a bit dangerous as if you have a start arg but no end, I'm not sure what should happen... Tuple is good, have altered

znicholls:
> Might be useful to elaborate or add a link to docs where this is explained a bit more.

Nice! Have done (although link won't work until docs are built...)

docs/file_conventions.rst
```diff
@@ -41,6 +41,28 @@ each of these files).
...
+- units: the units of the data
+
+- time: the point in time at which the data occurs
+
+    - note that the internal convention is that state variables are start of year values (i.e. 1st January 1990) whilst fluxes are annual averages/midyear values
+
+- set: a MAGICC flag which defines what MAGICC should do with the data in the timeseries
+
+    - SET: set the given variable in the given region to the value provided
+    - ADD: add the data to the given variable in the given region to any already read in data for that variable in that region (not clear when exactly this is taken into account, question to be answered)
```
rgieseke: Very useful to have this all written down -- just not clear to me what the comment in parens means ...?

docs/file_conventions.rst
```diff
@@ -41,6 +41,28 @@ each of these files).
...
+    - SUBTRACT: subtract the data to the given variable in the given region from any already read in data for that variable in that region (not clear when exactly this is taken into account, question to be answered)
```
rgieseke: subtract "from"?

znicholls:
@lewisjared hopefully this scenario builder should make your life a bit easier in future

pymagicc/io.py
```diff
@@ -2095,3 +2119,106 @@ def get_generic_rcp_name(inname):
...
+def _join_timeseries_mdata(base, overwrite, join_linear):
+    min_year = min(base["time"].min(), overwrite["time"].min())
+    max_year = max(base["time"].max(), overwrite["time"].max())
+    new_index = range(min_year, max_year + 1)
+
+    result = _reindex_df(base, new_index)
+    ow = _reindex_df(overwrite, new_index)
+
+    if result.stack().index.intersection(ow.stack().index).empty:
+        raise ValueError("No overlapping indices, a simple append will do")
```
lewisjared: Should this raise an exception? The function is meant to join time series so why not just return the appended dataframes in this case?

znicholls: My reasoning here was that if you call join_timeseries, you expect there to be some overwriting done and should be told if that's not the case. Maybe the function needs to be renamed to make that clearer?

znicholls: Or just raise a warning instead?

znicholls: It's to guard against the example in the notebook where you think you're joining scenarios, except something stupid has gone wrong (i.e. the todo column isn't behaving) and so you end up not joining anything and just getting a massive dataframe back which you can't use.

lewisjared: Ahh I see. That raises another question - what is the point of N/A in the TODO column if you just set it to set??? 

Perhaps, there needs to be a higher level function which takes scenario filenames and tries to merge them as it is a bit clunky having to load the scenarios and set the TODO column to set and then call join_timeseries

znicholls: > That raises another question - what is the point of N/A in the TODO column if you just set it to set???

I had this idea that cause they were data files, the user should have to choose what to do with them. On reflection, reading in SET makes more sense.

Will tweak and right high level option now

znicholls:
@lewisjared can you take another look?

lewisjared:
Sure if you push them :D

znicholls:
> Sure if you push them :D

fml

lewisjared:
LGTM once you resolve the failed test

PR #188, opened 2018-12-20
Make test-pypi-install also test imports of all dependencies

znicholls:
# Pull request

- [x] Tests added (N/A)
- [x] Documentation added (N/A)
- [x] Example added (N/A)
- [x] Description in ``CHANGELOG.rst`` added (N/A)

## Adding to CHANGELOG.rst

```
```

Just makes sure that when we import pymagicc, we go through and import all the dependencies in submodules too to double check that all the dependencies are importable (i.e. installed correctly)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ test-testpypi-install: venv
 		-i https://testpypi.python.org/pypi pymagicc \
 		--no-dependencies --pre
 	# Remove local directory from path to get actual installed version.
+	@echo "This doesn't test dependencies"
 	$(TEMPVENV)/bin/python -c "import sys; sys.path.remove(''); import pymagicc; print(pymagicc.__version__)"
 
 publish-on-pypi: venv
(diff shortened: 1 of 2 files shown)


rgieseke:
So this ensures that every (sub-)module can be imported without error? Shouldn't this be checked locally?

rgieseke:
It does work locally of course, just trying to better understand the use-case. Maybe it should also print something like "All submodules imported successfully", or just print "import pymagicc.io" and so on.

In any case it's good to have the check of the PyPI version in a separate file. One could also think about printing the PyPI version for comparison (but that's probably a separate issue).
It's available from their JSON API:
https://pypi.org/pypi/pymagicc/json

znicholls:
When I was developing netcdfscm, at first after a pip install the top level
import would work and I only realised that not all dependencies were
installing properly once I tried to import a submodule. This tweak just
made sure I would be alerted if the same issue were to arise again.

Happy to tweak printing and do pypi checks as you see fit

rgieseke:
I'm wondering if this should be part of the test suite ...

And also if after a `pip install` more or all tests should be run -- it's easy to miss some data files in MANIFEST or `setup.py` and then things don't actually work ...

znicholls:
With CI we could do a nightly pip install to check. Everytime doesn't work
I don't think cause you have to pip release before you know if it's a good
idea or not..

Running tests would force users to install all the test requirements, could
be undesirable..

rgieseke:
Alrighty! We can re-consider later (or create separate issues) ...

PR #192, opened 2018-12-26
WIP: Move to pyam backend

znicholls:
# Pull request

- [ ] Tests added
- [ ] Documentation added (where applicable)

## Adding to CHANGELOG.rst

```
```


znicholls:
@rgieseke @lewisjared this is pretty all over the place. Two things would be awesome:

1. high level comments on general premise of moving to pyam backend, I think my ideal would be to subclass `OpenSCMDataFrame` (which would live in openscm), with `OpenSCMDataFrame` itself being a subclass of `pyam.IamDataFrame`. This may be a bad idea, I just can't see why yet.
1. suggestions on how to break this into small enough steps to be useful for you guys, doing it all at once like this isn't going to work.

znicholls:
Addressed in commits which solve #197

PR #194, opened 2019-01-15
Update Linux version for Travis

rgieseke:


rgieseke:
> nice! is the check for python version 3.6 in lines 17&18 still necessary?

This is done to avoid running (the slow) notebook checks multiple times. We could probably add a comment.

PR #198, opened 2019-01-25
Tidy up requirements, all in setup.py now

znicholls:
# Pull request

- [x] Tests added (N/A)

## Adding to CHANGELOG.rst

```
```


rgieseke:
Good job! Can we pin expectexcept to a specific commit?

rgieseke:
Another question, `expectexceptation` is not really needed for running, it should go into tests, right?
`Appmode` and `ipywidgets` is only needed for running the Mybinder demo. We could also move it to a separate project (scm-demos?) ...

setup.py
```diff
@@ -48,19 +52,26 @@ def run_tests(self):
...
+    "expectexception @ git+https://github.com/thedataincubator/expectexception.git@master",
+     "ipywidgets",
+     "appmode",
 ]
 
-extra_requirements = {"test": ["pytest", "pytest-cov", "codecov", "goodtables"]}
+extra_requirements = {
+    "tests": ["pytest>=4.0", "pytest-cov", "codecov", "goodtables", "notebook", "matplotlib", "nbval", "goodtables"],
+    "docs": ["sphinx>=1.4", "sphinx_rtd_theme", "sphinx-autodoc-typehints", "pydoc-markdown"],
+    "deploy": ["setuptools>=38.6.0", "twine>=1.11.0", "wheel>=0.31.0", "flake8", "black", "versioneer"],
```
rgieseke: Maybe we should have a separate group `dev` or call the whole thing `dev` as only the first three are needed for PyPI.

Versioneer is not needed here i think, it's more a system tool that creates the necessary files in your project. Or is it needed to update `versioneer` from time to time?

znicholls:
> Can we pin expectexcept to a specific commit?

Yes. Might be a touch slow (https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip) but fine until one of us gets around making a request for a release in their repo.

znicholls:
> Another question, `expectexceptation` is not really needed for running, it should go into tests, right?
> `Appmode` and `ipywidgets` is only needed for running the Mybinder demo. We could also move it to a separate project (scm-demos?)

I thought about this and my feeling is that most people look at the notebooks when they use pymagicc. Hence all of the stuff someone needs to run the notebooks should be put as a dependency. Having people run notebooks and then get import errors just seems to be asking for trouble so I'd prefer to leave expectexception etc. in package dependencies...

setup.py
znicholls: > Maybe we should have a separate group `dev` or call the whole thing `dev` as only the first three are needed for PyPI.

Ok.

> Or is it needed to update `versioneer` from time to time?

Nope I don't think so, have removed.

rgieseke:
> Yes. Might be a touch slow (https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip) but fine until one of us gets around making a request for a release in their repo.

Interesting. For this small repo it shouldn't be a big problem - i also realize now that it's not such a big difference to other libs on PyPI as we also don't pin them.

It still feels clearer somehow :-)

znicholls:
> It still feels clearer somehow :-)

Haha whatever works

Issue #197, opened 2019-01-24
Move to using OpenSCMDataFrame


znicholls:
**Is your feature request related to a problem? Please describe.**

At the moment our interface is a bit clunky to use. We could make it much easier by moving to using OpenSCMDataFrame's in the back end instead.

**Describe the solution you'd like**

Move to using OpenSCMDataFrame throughout pymagicc.

**Describe alternatives you've considered**

Create our own implementation.

**Additional context**

znicholls:
@rgieseke @lewisjared I've attempted to address this in #198, #199, #200, #201, #202, #203 (they build on top of each other so hopefully all makes sense eventually).

There's a lot of dependencies which we're waiting on so the setup is a bit messier than would be ideal but hopefully these PRs should give us a start.

rgieseke:
This probably needs an update with the OpenSCM rename to `ScmDataFrame`

znicholls:
superceded by #295

PR #200, opened 2019-01-25
Intermediate: Pass io tests

znicholls:
Intermediate step towards solving #197. Fails CI as only the io tests have passed here.

@rgieseke the typing stuff sven has done in openscm mean that the code is now Python3.5 incompatible, are we ok with that?


rgieseke:
> @rgieseke the typing stuff sven has done in openscm mean that the code is now Python3.5 incompatible, are we ok with that?

Not sure what exactly the problem is, i thought `typing` was included in 3.5 - is a specific new feature the problem?

znicholls:
> Not sure what exactly the problem is, i thought `typing` was included in 3.5 - is a specific new feature the problem?

It appears not, from https://travis-ci.org/openclimatedata/pymagicc/jobs/484226040

```
E     File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/openscm/core.py", line 28
E       _world: _Region
E             ^
E   SyntaxError: invalid syntax
```

whilst https://travis-ci.org/openclimatedata/pymagicc/jobs/484226041 raises no such syntax error

rgieseke:
Hm, it seems this syntax is 3.6+ only.

Let's drop 3.5 support! If someone *needs* to run this on a server they can get a more recent Python with Conda.

pymagicc/io.py
```diff
@@ -242,20 +246,46 @@ def _convert_data_block_and_headers_to_df(self, stream, column_headers):
...
+            "climate_model",
+        ]:
             df[categorical_column] = df[categorical_column].astype(str)
 
         return df
 
     def _get_columns_multiindex_from_column_headers(self, ch):
+        l = len(ch["variables"])
+        ch.setdefault("climate_models", ["MAGICC"] * l)
+        ch.setdefault("models", ["tbc"] * l)
```
rgieseke: Is "tbc" justs a placeholder or does it need to be changed later?

znicholls:
placeholder, changed to "unspecified" in #201 or #202

PR #208, opened 2019-01-31
Document MAGICC run mode flags

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


znicholls:
@rgieseke this is good to, let me know if you want me to split it into smaller pieces!

The fix for appveyor failure is to re-authorise apparently https://help.appveyor.com/discussions/problems/11287-the-build-phase-is-set-to-msbuild-mode-default-but-no-visual-studio-project-or-solution-files-were-found

znicholls:
start with this, it's the most interesting https://github.com/openclimatedata/pymagicc/blob/252c0c1d8c837fd78f97a46c2270b4b9283918b5/notebooks/Run-Modes.ipynb

znicholls:
@rgieseke can I push you on this one?

rgieseke:
> @rgieseke can I push you on this one?

Sure!

Quick questions, this is mixing in a few different things … the documentation of the run modes and forcing only is super awesome to have!

Why is `read_scen_file` in `scenarios` module, not in `io`?

I find the extra IAM model column a bit confusing, while the RCPs have been generated with these i think their meaning is not really connected to them anymore for the purposes of running an RCP scenario.

In the tests, what do you mean by a "package"?

Otherwise, i think this can be merged!

znicholls:
> this is mixing in a few different things

Ye for sure, can split too if you want.

> Why is `read_scen_file` in `scenarios` module, not in `io`?

Good question. Originally it was in `__init__.py` and then I moved it but made sure it was still accessible via `__init__.py` so wasn't really thinking. Will move it

rgieseke:
> Me too. My thoughts here were that we could either sometimes have that column and sometimes not or always have. I lean towards always having it as pyam distinguishes emissions pathways with the combination of model and scenario so you do kind of need that extra column to get a unique identifier (there are e.g. RCP emissions generated with other models, they're just not 'the RCPs'). ScmDataFrame doesn't require these columns so we don't necessarily need them, I'm just always in favour of having these things always there, rather than having them appear unexpectedly later.

Can we have them as `None` and don't display if `is None`?

rgieseke:
> Ye for sure, can split too if you want.

Next time ;-) No need to cherry pick ...

znicholls:
> Can we have them as `None` and don't display if `is None`?

If we don't display we should just get rid of them (the point of keeping them is to remind the user that they're there). Getting rid of this column also makes the translation to iam_dataframe slightly harder (you always have to add a model column when you do translate to IamDataFrame).

rgieseke:
> If we don't display we should just get rid of them (the point of keeping them is to remind the user that they're there). Getting rid of this column also makes the translation to iam_dataframe slightly harder (you always have to add a model column when you do translate to IamDataFrame).

In the plots? In the DataFrame they can easily be dropped.

znicholls:
> In the plots? In the DataFrame they can easily be dropped.

Where are you talking about? In an IamDataFrame you can't drop the model column.

rgieseke:
> Where are you talking about? In an IamDataFrame you can't drop the model column.

I mean, when i want to further analyse, reshape the `.df` i can easily do so (and users do this all the time when working with DataFrames) -- for a default plot it could be done in the plotting function without needing user interaction.

But this is really something for fine-tuning later. (If model is `None`, don't put it in the title string)

znicholls:
Ah ok but are you just talking about plotting, or also about what we read
in (ie do you not want to include the model name when reading in the rcps?)?

rgieseke:
> Ah ok but are you just talking about plotting, or also about what we read in (ie do you not want to include the model name when reading in the rcps?)?

If it's in the SCEN file we could as well read it in of course.
I just find it confusing when an it shows up as unspecified in a plot, in that case it could just be left out, couldn't it?

image

znicholls:
> I just find it confusing when an it shows up as unspecified in a plot, in that case it could just be left out, couldn't it?

Oh that. Yes in theory easy to not plot, in practice harder because all the plotting code is in pyam (but still not that hard). How much do you want it removed?

rgieseke:
> Oh that. Yes in theory easy to not plot, in practice harder because all the plotting code is in pyam (but still not that hard). How much do you want it removed?

No hurry, just thought it might be easier if the value is `None` vs. having to check against
"Unspecified" or different spellings people might accidentally use.

So instead of 

```python
columns={
        "index": time,
        "scenario": ["idealised"],
        "model": ["unspecified"],
        "climate_model": ["unspecified"],
```

have something like
```python
columns={
        "index": time,
        "scenario": ["idealised"],
        "model": [None],
        "climate_model": [None],
```
Or does that need to be a string?

znicholls:
> Or does that need to be a string?

I don't believe so. I think moving to `None` is the right call! I'll try it in here and see how much chaos it causes (thanks for your patience with OpenScm moving all over the shop)

znicholls:
> I think moving to `None` is the right call! I'll try it in here and see how much chaos it causes (thanks for your patience with OpenScm's `scmcallib` branch moving all over the shop)

Thinking again, let's address this in OpenScm? (see https://github.com/openclimatedata/openscm/issues/114)

PR #213, opened 2019-02-01
Update setup.py

znicholls:
Damn it I should have read the issues first. 

https://bugs.python.org/msg93286 gives a solution kind of like this.
Maybe making a shared email address is a good idea...


rgieseke:
Don't think that's a good solution - one place more where we have to keep email adresses updated. I'd suggest to stick with the current version (from my reading of these threads it doesn't matter and doesn't need to be machine readable) or to have only your contact info there, as technically, you are the current maintainer :-)

znicholls:
> only your contact info there

haha no

PR #216, opened 2019-02-12
change index in tests as years are cast to datetimes

lewisjared:
@znicholls This should resolve the failing tests


znicholls:
@rgieseke I've pushed this one through as it's tiny, hope that's all good

lewisjared:
@znicholls it should pass this time. Obviously, ScmTimeSeries can't guarantee to preserve meta row ordering.

tests/test_io.py
```diff
@@ -2614,6 +2614,10 @@ def test_magicc_data_append(mock_read_and_return_metadata_df):
     expected_metadata = deepcopy(tmetadata_init)
     expected_metadata.update(tmetadata_append)
     assert mdata.metadata == expected_metadata
+
+    # update the tdf_init index as year values are converted to datetimes
+    tdf_init.index = mdata.timeseries().columns
+    tdf_append.index = mdata.timeseries().columns
     pd.testing.assert_frame_equal(
         mdata.timeseries().reset_index(drop=True),
```
znicholls:
```suggestion
        mdata.timeseries().sort_index().reset_index(),
```

tests/test_io.py
```diff
@@ -2614,6 +2614,10 @@ def test_magicc_data_append(mock_read_and_return_metadata_df):
...
         tdf_init.T.append(tdf_append.T).reset_index(drop=True),
```
znicholls:
```suggestion
        tdf_init.T.append(tdf_append.T).sort_index().reset_index(),
```

lewisjared: tdf_init.T has a junk index so sorting didn't work. Just needed to explicitly sort mdata.timeseries() as that the one which was inconcistent

lewisjared: Also needed the drop since the indexes aren't the same

lewisjared:
@znicholls about time

znicholls:
hang on I think we can make this a bit more obvious

znicholls:
I'll make a PR now

lewisjared:
@znicholls Good to go?

PR #220, opened 2019-02-15
Prioritise binout over out files

lewisjared:
# Pull request

- [x] Tests added

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ master
 
 - (`#223 <https://github.com/openclimatedata/pymagicc/pull/223>`_) Ensure `pymagicc.io._BinaryOutReader` closes the input file
 - (`#222 <https://github.com/openclimatedata/pymagicc/pull/222>`_) Remove trailing `/` in `MANIFEST.IN` recursive includes as this is invalid syntax on windows.
+- (`#220 <https://github.com/openclimatedata/pymagicc/pull/220>`_) If binary and ascii output files exist for a given variable only read the binary file
 - (`#208 <https://github.com/openclimatedata/pymagicc/pull/208>`_) Add set zero config method. Also adds scenarios module, tidies up the notebooks and adds a notebook showing how to run in different modes.
 - (`#214 <https://github.com/openclimatedata/pymagicc/pull/214>`_) Refactor to use the timeseries capabilities of ScmDataFrameBase
 - (`#210 <https://github.com/openclimatedata/pymagicc/pull/210>`_) Updated to match new openscm naming
--- a/pymagicc/core.py
+++ b/pymagicc/core.py
@@ -225,7 +225,7 @@ class MAGICCBase(object):
         # On Windows shell=True is required.
         subprocess.check_call(command, cwd=self.run_dir, shell=IS_WINDOWS)
 
-        outfiles = [f for f in listdir(self.out_dir) if f != "PARAMETERS.OUT"]
+        outfiles = self._get_output_filenames()
 
         read_cols = {"climate_model": ["MAGICC{}".format(self.version)]}
         if scenario is not None:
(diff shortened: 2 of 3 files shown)


znicholls:
pros:
- speed
- leaves unit handling to openscm (when ready)

cons:
- prioritises files which don't have units

lewisjared:
I wasn't able to read SLR_EXPANSION correctly with master as the different units/TODO values do not merge
```
from pymagicc import MAGICC6

with MAGICC6() as m:
    m.set_output_variables(write_binary=True, sealevel=True)
    result = m.run(only=('SLR_EXPANSION',))
    print(result.timeseries())
```
```
time                                                                                              1765-01-01  ...  2100-01-01
model       scenario    region                          variable      unit    climate_model todo              ...            
unspecified unspecified World                           SLR_EXPANSION cm      MAGICC6       N/A          0.0  ...   20.490638
                        World|Northern Hemisphere|Land  SLR_EXPANSION cm      MAGICC6       N/A          0.0  ...   20.490638
                        World|Northern Hemisphere|Ocean SLR_EXPANSION cm      MAGICC6       N/A          0.0  ...   20.490638
                        World|Southern Hemisphere|Land  SLR_EXPANSION cm      MAGICC6       N/A          0.0  ...   20.490638
                        World|Southern Hemisphere|Ocean SLR_EXPANSION cm      MAGICC6       N/A          0.0  ...   20.490638
                        World                           SLR_EXPANSION unknown MAGICC6       SET          0.0  ...   20.490638
                        World|Northern Hemisphere|Land  SLR_EXPANSION unknown MAGICC6       SET          0.0  ...   20.490638
                        World|Northern Hemisphere|Ocean SLR_EXPANSION unknown MAGICC6       SET          0.0  ...   20.490638
                        World|Southern Hemisphere|Land  SLR_EXPANSION unknown MAGICC6       SET          0.0  ...   20.490638
                        World|Southern Hemisphere|Ocean SLR_EXPANSION unknown MAGICC6       SET          0.0  ...   20.490638
```

lewisjared:
I think that it is safe to always prioritise DAT_*.BINOUT over DAT_*.OUT
I'll update this to ignore ocean layer binaries as they can't currently be read

znicholls:
> I wasn't able to read SLR_EXPANSION correctly with master as the different units/TODO values do not merge

ughhh what, how does the underlying data file header look?

lewisjared:
The .OUT file is correct with units available in the header, but it shows that the .OUT and .BINOUT data can't be merged - therefore you should only read one. In this case, I need both ascii and binary output as there currently isn't a TEMP_OCEANLAYER.BINOUT reader, but I still would also like the SLR_EXPANSION variable.

If you have turned on binary output (isn't the default) you're trying to speed up the execution time. If you read both files formats it is only ever going to be slower than a single format, completely negating the need for reading the binary files.

znicholls:
> therefore you should only read one

I ok I get it now. Why do we prioritise BINARY though? Cause it's faster? If you're writing both you want pymagicc to be careful (i.e. read ASCII) than fast (if you wanted fast, you'd run BINARY only)? Or am I missing the point here?

rgieseke:
Awesome! Shouldn't binary output be the default (if it's faster) for a session where the temp dir gets auto-deleted?

lewisjared:
I'd still argue that if binary output has been turned on it should always be preferenced (if it can be read) because writing binary files it isn't the default state. A configuration change had to be made to write those binary files which suggests that the user is happy with the performance/unit tradeoff.

rgieseke:
But where is the trade-of if a user never sees the temp dir? Whether the unit "definition" is done in the output file or in Pymagicc code doesn't matter, does it?

lewisjared:
> But where is the trade-of if a user never sees the temp dir?

That is true. The question becomes are ascii + binary formats written or just binary. If just binary, you would have to explicitly turn ascii on if you wanted ocean layer temperatures. I'm not sure if there are any other edge cases, perhaps for some of the more exotic variables, there are also only ascii readers. If that is the case then as a user I would be happy with binary-only for 99% of the time.

> Whether the unit "definition" is done in the output file or in Pymagicc code doesn't matter, does it?

No it doesn't matter where the units come from. It does add a burden of maintaining a list of units to pymagicc.

lewisjared:
The failing appveyor build was from an old version of openscm, but I can't force a rerun without adding another commit

lewisjared:
@rgieseke Happy for this to be merged and for temp directories defaulting to binary output to be implemented in a separate PR?

PR #223, opened 2019-02-20
Ensure input file is closed

lewisjared:
# Pull request

- [ ] Tests added

I wasn't able to produce a simple test to reproduce the output seen when running scmcallib, but I think that this resolves the error message. Anyway it is better practise to use a context manager with open file handles

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 master
 ------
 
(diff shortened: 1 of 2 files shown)


znicholls:
> I wasn't able to produce a simple test to reproduce the output seen when running scmcallib, but I think that this resolves the error message

I don't think that's an issue, as you say the context manager should cover it. Thank you!

PR #225, opened 2019-02-27
Fix permafrost read in

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

### DIFF
--- a/pymagicc/definitions/__init__.py
+++ b/pymagicc/definitions/__init__.py
@@ -251,6 +251,9 @@ def get_magicc7_to_openscm_variable_mapping(inverse=False):
         if "CO2CH4N2O" in variable:
             variable = variable.replace("CO2CH4N2O", "CO2, CH4 and N2O")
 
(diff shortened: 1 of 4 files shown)


pymagicc/definitions/__init__.py
```diff
@@ -251,6 +251,9 @@ def get_openscm_replacement(in_var):
         if "CO2CH4N2O" in variable:
             variable = variable.replace("CO2CH4N2O", "CO2, CH4 and N2O")
 
+        if "PF" in variable:
```
lewisjared: I'm assuming that there is only one PF variable? Can this be made more specific to avoid any unintended side effects

znicholls: That's a good point. Looking at them now, the PF variables are going to be super hard to do more generally so I've moved this into handling that case specifically. I'll leave a more general permafrost handling algorithm for another time.

PR #227, opened 2019-03-07
Tweak permafrost read

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

This changes the way permafrost emissions are named. Putting them under emissions is problematic as emissions are implicitly anthropogenic. If you have "Emissions|CO2|Permafrost" then when you add up all sub-categories of CO2 you end up with non-anthropogenic emissions in there too which isn't what you want. This renaming should avoid that sort of thing happening and allows Emissions to remain being implicitly anthropogenic (unfortunately we shouldn't make it explicit as then things will explode in terms of mapping to pyam variables where emissions are always anthropogenic).

### DIFF
--- a/pymagicc/definitions/__init__.py
+++ b/pymagicc/definitions/__init__.py
@@ -340,8 +340,8 @@ def get_magicc7_to_openscm_variable_mapping(inverse=False):
             "CH4OXSTRATH2O_RF": "Radiative Forcing|CH4 Oxidation Stratospheric H2O",  # what is this
             "LANDUSE_RF": "Radiative Forcing|Land-use Change",
             "BCSNOW_RF": "Radiative Forcing|Black Carbon on Snow",
-            "CO2PF_EMIS": "Emissions|CO2|MAGICC Permafrost",
-            "CH4PF_EMIS": "Emissions|CH4|MAGICC Permafrost",
+            "CO2PF_EMIS": "Land to Air Flux|CO2|MAGICC Permafrost",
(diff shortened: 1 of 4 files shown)


pymagicc/definitions/__init__.py
```diff
@@ -340,8 +340,8 @@ def get_openscm_replacement(in_var):
             "CH4OXSTRATH2O_RF": "Radiative Forcing|CH4 Oxidation Stratospheric H2O",  # what is this
             "LANDUSE_RF": "Radiative Forcing|Land-use Change",
             "BCSNOW_RF": "Radiative Forcing|Black Carbon on Snow",
-            "CO2PF_EMIS": "Emissions|CO2|MAGICC Permafrost",
-            "CH4PF_EMIS": "Emissions|CH4|MAGICC Permafrost",
+            "CO2PF_EMIS": "Land to Air Flux|CO2|MAGICC Permafrost",
+            # "CH4PF_EMIS": "Land to Air Flux|CH4|MAGICC Permafrost",
```
rgieseke: Why commented out now?

znicholls: cause I haven't tested it, I can also just remove if you want

rgieseke: Dunno, either add "TODO: Add when tested or needed" or something, or remove it to avoid confusion ...

znicholls:
```suggestion
            # "CH4PF_EMIS": "Land to Air Flux|CH4|MAGICC Permafrost",  # TODO: test and then add when needed
```

PR #229, opened 2019-04-08
Add more robust test of io

znicholls:
# Pull request

- [x] Tests added
- [x] Example added (either to an existing notebook or as a new notebook, where applicable) (N/A, all internal)

## Adding to CHANGELOG.rst

```
```


znicholls:
@lewisjared can you take a look when you've got a minute? Let's not merge until @rgieseke gives the approval though

pymagicc/io.py
```diff
@@ -807,17 +807,22 @@ def read(self):
...
-            elif metadata["unit"] == "metric tons":
-                emms = True
-            else:
-                error_msg = "I do not recognise the unit, {}, in file{}".format(
-                    metadata["unit"], self.filepath
-                )
-                raise ValueError(error_msg)
-        else:
+
+        unit_keys = ["Unit", "unit"]
```
lewisjared: Why not perform a case insensitive lookup (using tolower())?

pymagicc/io.py
```diff
@@ -872,13 +877,14 @@ def _read_header(self):
...
+        data_block_header_line = self._stream.readline().replace("Years", "").strip()
         variables = []
         col_width = 10
-        for n in np.arange(0, len(data_block_header_line), col_width):
-            pos = int(n)
-            variable = data_block_header_line[pos : pos + col_width].strip()
+
+        for w in range(0, len(data_block_header_line), col_width):
+            variable = data_block_header_line[w : w + col_width].strip()
             if variable != "Years":
```
lewisjared: You have already stripped out "Years" in L882

znicholls:
> Are any of these expected files actually run through magicc to be sure that they work?

They are now (except the OT file because I have no idea how that's used and hence how to test it...)

znicholls:
@rgieseke if you have time would be great to get your thoughts

pymagicc/io.py
```diff
@@ -807,20 +807,27 @@ def read(self):
...
-                error_msg = "I do not recognise the unit, {}, in file{}".format(
-                    metadata["unit"], self.filepath
+
+        unit_keys = [k for k in metadata.keys() if k.lower() == "unit"]
+        units = [metadata.pop(k) for k in unit_keys]
+        if not units:
+            # have to assume global emissions in tons
+            emms = True
+        elif len(units) != 1:
+            import pdb
```
rgieseke: Looks super useful on one hand, but could be very irritating to users not used to Python debugging ...

rgieseke: Or is this a left-over?

rgieseke:
Excellent documentation of pit-falls, very useful to have this written down here!
Only question from my side the debug statement ...

pymagicc/io.py
znicholls: left over, good pick up

znicholls:
@lewisjared over to you to merge when you're happy then I think

PR #239, opened 2019-04-18
Override magicc7 user config with sane defaults

lewisjared:
# Pull request

The MAGCFG_USER.CFG file in the magicc repository is in a constant state of flux. It would be nice to overwrite the MAGICC7 config in pymagicc rather than duplicating this everywhere

- [x] Tests added

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 master
 ------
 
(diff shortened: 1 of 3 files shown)


lewisjared:
This has a byproduct of making (most) of the MAGICC7 tests pass

PR #246, opened 2019-05-13
WIP: Add demo notebook for global/regional scenarios

rgieseke:
This is a simple example trying to compare regionally split RCPs with a globally aggregated scenario.
Resulting global temperature differences are never > 0.01.


znicholls:
Nice. I'd split the running code from the analysis (finding the max) and the plotting code but then I think we're good!

rgieseke:
>Nice. I'd split the running code from the analysis (finding the max) and the plotting code but then I think we're good!

Done. See https://github.com/openclimatedata/pymagicc/blob/18adb3f71c9a3c5514e06110d2540f459513851c/notebooks/Regional-Split.ipynb

Pinging @gidden and @danielhuppmann who might be interested

rgieseke:
Thanks @znicholls - that's excellent explanation text. I've kept the sentence from the 2nd cell and included it. I think it's nice to have the summary somehow because all these numbers and graphs could look overwhelming. What do you think?

znicholls:
Nice! I'd change the title to "Comparing Regionally Split and Global Scenarios".

Otherwise happy to merge when you pull of the "WIP" label

znicholls:
@rgieseke ping I think this is good to go so merge away!

Issue #166, opened 2018-10-21
Discussion: tidy up post 2.1.0


znicholls:
So I think for 2.1.0 we should aim to:

- [ ] integrate with openscm 
- [ ] use of some variant of pyam dataframe

Then I think we should focus on refactoring things, in particular:

- [ ] tidying up docs
- [ ] re-write any untidy code
- [ ] split the tests into unit and integration so it becomes easier to see what's going on and develop in future

znicholls:
yes but if we break that API (i.e. rename `pymagicc.api` to `pymagicc.core`) then we should really bump to 3.0 if we're going to follow semantic versioning properly. Plan b is we change it before the release?

rgieseke:
Both possible, just wanted to note down the idea. We can also provide an alias and deprecation warning but that's probably overkill given we have hardly any users (and need to fix the MAGICC7 pipeline ourselves).

PR #249, opened 2019-06-07
MAGICC7 updates

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

@rgieseke if you checkout the pymagicc-2-0-running branch of magicc and do `export MAGICC_EXECUTABLE_7=path/to/magicc/bin/magicc` then do `make test` all the tests should pass...

### DIFF
--- a/pymagicc/core.py
+++ b/pymagicc/core.py
@@ -768,8 +768,6 @@ class MAGICCBase(object):
             the experiment i.e. atmospheric |CO2| concentrations, total radiative
             forcing and global-mean surface temperature
         """
-        if self.version == 7:
-            raise NotImplementedError("MAGICC7 cannot yet diagnose ECS and TCR")
         self._diagnose_tcr_ecs_config_setup(**kwargs)
         timeseries = self.run(
             scenario=None,
@@ -789,8 +787,10 @@ class MAGICCBase(object):
 
         self.update_config(
             FILE_CO2_CONC="TCRECS_CO2_CONC.IN",
+            CO2_SWITCHFROMCONC2EMIS_YEAR=30000,
             RF_TOTAL_RUNMODUS="CO2",
             RF_TOTAL_CONSTANTAFTERYR=2000,
+            # CORE_CO2CH4N2O_RFMETHOD="IPCCTAR",
             **kwargs,
         )
 
@@ -866,6 +866,8 @@ class MAGICCBase(object):
 
     def _check_tcr_ecs_total_RF(self, df_total_rf, tcr_time, ecs_time):
         total_rf = df_total_rf.timeseries()
+        total_rf_max = total_rf.values.squeeze().max()
+
         t_start = total_rf.columns.min()
         t_end = total_rf.columns.max()
         t_start_rise = tcr_time - relativedelta(years=70)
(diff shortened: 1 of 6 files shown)


znicholls:
> I'm not a huge fan of the tcr functionality sitting in the MAGICCBase class

The tcr diagnosis functionality? Yes I agree, let's do it in #166

znicholls:
@rgieseke over to you for final review and merge

pymagicc/core.py
```diff
@@ -149,24 +154,111 @@ def create_copy(self):
...
+                    _deep_update(b[k], v)
+                else:
+                    b.update(o)
+
+        _deep_update(self._default_config, user)
+
+        return self._default_config
+
+    @property
+    def fgas_files_conc(self):
```
rgieseke: Is this a new flag in MAGICC7?

pymagicc/core.py
```diff
@@ -149,24 +154,111 @@ def create_copy(self):
...
+
+        Returns
+        -------
+        str
+            Name of the flag
+        """
+        return "fgas_files_conc"
+
+    @property
+    def mhalo_switchfromconc2emis_year(self):
```
rgieseke: Same here ... why is this a separate property?

pymagicc/core.py
znicholls: The flag name changed between MAGICC6 and 7. Adding the property means the code doesn’t have to worry about that change.

pymagicc/core.py
znicholls: As above

znicholls:
Hopefully that makes sense now @rgieseke

rgieseke:
> The flag name changed between MAGICC6 and 7. Adding the property means the code doesn’t have to worry about that change.

But is it still possible to change these values like all the others?

znicholls:
> But is it still possible to change these values like all the others?

Yes (of course if you set a flag that doesn't exist then MAGICC will blow up so if you're going to do it by hand you have to be more careful about which MAGIC you use))

rgieseke:
So why not have this handled in `set_config`? (Maybe i'm missing something, but having these extra properties looks somehow wrong to me ...)

znicholls:
> So why not have this handled in `set_config`?

That would also be a solution which I'm happy to implement (tomorrow). I think I thought properties seemed like a good idea in case I ever needed to query the class for this flag later.

tests/test_core.py
```diff
@@ -679,11 +676,12 @@ def test_persistant_state(package):
 
 
 def test_persistant_state_integration(package):
-    if package.version == 7:
-        pytest.xfail(reason="MAGICC7 TCR/ECS diagnosis is currently broken")
-        package.diagnose_tcr_ecs()
     test_ecs = 1.75
     package.update_config(CORE_CLIMATESENSITIVITY=test_ecs)
+    if package.version == 7:
+        # ecs only follows its definition with IPCCTAR forcing (MAGICC7 bug...)
```
rgieseke: I'm a bit worried that this might go out of date (once it's fixed) - is there an issue number which could be tracked (even if in private repo)?

tests/test_core.py
```diff
@@ -807,22 +809,42 @@ def test_pymagicc_writing_has_an_effect(
...
         # special undocumented flags!!!
         relevant_config["fgas_adjstfutremis2past_0no1scale"] = 0
         relevant_config["mhalo_adjstfutremis2past_0no1scale"] = 0
 
-    for key, value in relevant_config.items():
+    iter_dict = copy.deepcopy(relevant_config)
+    for key, value in iter_dict.items():
         if value == "test_filename":
             relevant_config[key] = test_filename
+        if key == "file_mhalo_emis" and package.version == 7:
```
rgieseke: Great! Maybe add a comment like "Handle parameters renamed in MAGICC7"?

tests/test_core.py
znicholls: Ok I re-wrote the test to make it tracked here. See what you think

znicholls:
@rgieseke if you're happy can you merge (I don't like merging if I authored the PR)

rgieseke:
@znicholls Sorry, missed that you pushed more commits!

PR #250, opened 2019-06-12
Add support for `.MAG` files

znicholls:
# Pull request

`.MAG` files are the format we're aiming to re-write all of MAGICC's input and output files in (eventually). This PR adds support for this reading and writing to help facilitate the move.

- [x] Tests added
- [x] Merged #249 (this one follows straight on)

## Adding to CHANGELOG.rst

```
```


znicholls:
should follow #249

docs/file_conventions.rst
```diff
@@ -40,6 +38,10 @@ each of these files).
    deprecated for MAGICC7 in favour of ``*EMIS.IN`` and ``*.SCEN7``
    files.
 
+In future, the MAGICC developers are aiming to move all of MAGICC's input and output
```
rgieseke: For future reference it might be useful to add some of the thinking about the new format here ... e.g. describe the blocks.

znicholls: Something like, “this format will always be comprised of 4 sections. The first is a header, for storing whatever text the user wants. This section must always start with ‘—- Header —-‘. It is ignored by MAGICC but can be used by other readers and writers. The next section is other metadata, in key:value pairs. This section must always start with ‘—- Metadata —-‘. The section allows for storage of metadata, like the global attributes section in a netcdf file. This section is also ignored by MAGICC. The third section is a FORTRAN namelist, which stores the flags required to read the file in MAGIC. The flags must match the flags used by MAGICC internally [add link to namelist section below here]. In Pymagicc, these flags are written automatically, the user can not write them. The fourth section is the data. This is always a data block with four header rows: variable, todo, units, region [TODO link to meaning of rows] (although the region row label is ‘Years’). The first column is the time axis and the subsequent columns are the timeseries. This format is highly custom and specialised for use with MAGICC but also to allow human readability. If you want to work with the data, we strongly recommend using Pymagicc’s ‘io’ module to allow easy conversion to python types such as dictionaries, lists, strings and most importantly pandas data frames.”

znicholls: ?

rgieseke: Great, that clarifies a lot for me!

Is the intention to have something that is "better" specified as previous formats but still readable easily with minimal changes to the Fortran MAGICC readers?

znicholls:
Ye basically. Although I think we’ve tried to focus on better specified and
useful, it just happens to be that the solution we’ve come up with is not a
huge change.

docs/file_conventions.rst
znicholls: In short, yes. Focus a) was definitely on something that was better specified. The fact it's similar to previous stuff is (in my opinion) more luck than design. If it had needed to be completely different, we probably would have changed it. It's just that MAGICC only really has timeseries data like this so the tabular format is fine.

znicholls:
@lewisjared ping does this fit with your understanding of `.MAG` files?

pymagicc/io.py
```diff
@@ -1445,17 +1514,20 @@ def _convert_data_block_to_magicc_time(self, data_block):
         else:
 
             def _convert_to_decimal_year(itime):
-                # MAGICC dates are to nearest month at most precise
+                # MAGICC dates are either middle of start of month
```
lewisjared: "either middle or start of month"

tests/test_io.py
```diff
@@ -44,8 +44,8 @@
 
 INVALID_OUT_FILES = [
     r"CARBONCYCLE.*OUT",
-    r".*SUBANN.*OUT",
-    r"DAT_VOLCANIC_RF\..*OUT",
+    r".*SUBANN.*\.BINOUT",
```
lewisjared: Is this change to ensure that the .OUT files are still readable. I'm fairly sure that we exclude them in the actual reader.

pymagicc/io.py
```diff
@@ -1445,17 +1514,20 @@ def _convert_data_block_to_magicc_time(self, data_block):
...
-                hr = itime.hour
-                month_fraction = (day + hr / 24) / month_days
-                # decide if midyear or start of year
-                if np.round(month_fraction, 1) == 0.5:
+
+                year_fraction = (itime - datetime(year, 1, 1)).total_seconds() / (
+                    datetime(year + 1, 1, 1) - datetime(year, 1, 1)
+                ).total_seconds()
+
+                startmonths = np.arange(0, 1, 1 / 12)
```
lewisjared: So magicc uses a 360 day calendar?

tests/test_data/MAG_FORMAT_SAMPLE.MAG
```diff
@@ -0,0 +1,49 @@
+---- HEADER ----
+
```
lewisjared: Are the empty lines part of the spec?

tests/test_io.py
znicholls: It's because we can read subannual `.OUT` files now. I deleted that restriction about annual timesteps in the ASCII readers (see https://github.com/openclimatedata/pymagicc/pull/250/files#diff-1e016ee4725860a3cd5d802fca30655bL194). The test to make sure this is all good is https://github.com/openclimatedata/pymagicc/pull/250/files#diff-cc4345db19ff44863e9122c74e9f383fR1170 (it checks that it's on a monthly grid and that annual times are still read as desired)

pymagicc/io.py
znicholls: MAGICC doesn't have an internal concept of days so I wouldn't say it's a 360 day calendar. In #252 I phrase it as 'MAGICC has a year, divided into 12 equal length months'

tests/test_data/MAG_FORMAT_SAMPLE.MAG
znicholls: Nope empty lines are ignored on read in but there's always empty lines added on writing.

lewisjared:
I'll let @rgieseke merge after he's had a review

docs/file_conventions.rst
```diff
@@ -12,10 +12,8 @@ them, they will a) run MAGICC or b) be handled in the expected way.
 The files in the ``run`` directory follow a number of conventions, we
 summarise these here (please note that this is just a summary, we are
 still unsure about how strictly many of these conventions have to be
-followed as we are yet to set up tests for running MAGICC with all of
-these different files after they’ve been set using pymagicc, the key
-question we still need to answer is how important column order is in
-each of these files).
+followed as we are yet to set up tests for running MAGICC with *all* of
+these different files after they’ve been set using pymagicc).
```
rgieseke:
```suggestion
these different files after they’ve been set using Pymagicc).
```

docs/file_conventions.rst
```diff
@@ -129,3 +126,38 @@ write a ``SCEN7`` file and we are given the RCP regional set, will simply assume
...
+The first section of the file is a header, for storing whatever text the user wants.
+This section must always start with ``---- HEADER ----``. The section is ignored by
+MAGICC but can be used by other readers and writers.
+
+The next section is other metadata, in ``"key: value"`` pairs. As a result, each key
+and value can only be stored and read as strings. This section must always start with
+```---- METADATA ----```. The second section allows for storage of metadata, like the
+global attributes section in a netcdf file. This section is also ignored by MAGICC.
+
+The third section is a FORTRAN namelist, which stores the flags required for MAGICC to
```
rgieseke:
```suggestion
The third section is a Fortran namelist, which stores the flags required for MAGICC to
```

docs/file_conventions.rst
```diff
@@ -129,3 +126,38 @@ write a ``SCEN7`` file and we are given the RCP regional set, will simply assume
...
+write them.
+
+The fourth section is the data. This is always a data block with four header rows:
+variable, todo, units, region (see `Column meaning`_). In the data block, the first
+column is the time axis and the subsequent columns are the timeseries.
+
+This format is highly custom and specialised for use with MAGICC, with the secondary
+characteristic of being somewhat human readable. Having said this, if you want to work
+with the data, we strongly recommend using Pymagicc's io module (:ref:`pymagicc.io`) to
+allow easy conversion to more familiar python types such as dictionaries, lists,
```
rgieseke:
```suggestion
allow easy conversion to more familiar Python types such as dictionaries, lists,
```

pymagicc/io.py
```diff
@@ -1008,6 +1013,50 @@ def _read_magicc6_style_header(self, stream, metadata):
...
+            elif line == "---- METADATA ----":
+                in_header = False
+            else:
+                if in_header:
+                    header_lines.append(line)
+                else:
+                    try:
+                        key, value = line.split(":")
+                    except ValueError:
+                        raise ValueError(  # pragma: no cover # emergency valve
```
rgieseke: What's the reason for raising the excepted error?

znicholls: Make the error message clearer

znicholls:
> I think one could have been bolder for something even more modern but if this replaces the zoo of file formats that's already great progress!

Definitely but the negotiating team weren't strong enough to get that through. I was thinking of suggesting datapackage type files, what would be your preffered format (if you could choose anything)?

znicholls:
@rgieseke or @lewisjared I think this is now gtg

rgieseke:
>  I was thinking of suggesting datapackage type files, what would be your preffered format (if you could choose anything)?

I've been thinking about this quite a bit recently. Data Packages definitely fall short on the human readable side ... and there is compromising on building the reader backend in the Fortran side.
Now it's well tested and documented, so progress!

pymagicc/io.py
rgieseke: Sorry @znicholls , what i meant was why not just print the error message:

```python
line = "no colon"
try:
    key, value = line.split(":")
except ValueError:
    print(  # pragma: no cover # emergency valve
      "Something wrong, line should be of the form 'key: value' "
      "with no other colons (':').\n"
      "Problem line: {}".format(line)
)
```

Output
```
Something wrong, line should be of the form 'key: value' with no other colons (':').
Problem line: no colon
```

Vs.
```python
line = "no colon"
try:
    key, value = line.split(":")
except ValueError:
    raise ValueError(  # pragma: no cover # emergency valve
      "Something wrong, line should be of the form 'key: value' "
      "with no other colons (':').\n"
      "Problem line: {}".format(line)
)
```

Output with both exceptions:
```
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    key, value = line.split(":")
ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test.py", line 8, in <module>
    "Problem line: {}".format(line)
ValueError: Something wrong, line should be of the form 'key: value' with no other colons (':').
Problem line: no colon
```

znicholls: It doesn’t raise in that case does it? That could be better, I hadn’t actually thought it through

znicholls: I think I changed it in #252 anyway so suggestions welcome there!

PR #252, opened 2019-06-17
Improve header writing, add magicc_time module and add better wine not installed message

znicholls:
# Pull request

- [x] Tests added
- [x] Merge #250 

## Adding to CHANGELOG.rst

```
```


docs/file_conventions.rst
```diff
@@ -15,22 +17,32 @@ still unsure about how strictly many of these conventions have to be
...
+     Specifically, when a ``.SCEN`` file is written, ``writer.metadata["description"]``
+     is written in the fourth line of the file, ``writer.metadata["notes"]`` is
+     written in the fifth line of the file and all other metadata is written at the
+     end of the file. If ``description``/``notes`` is not available in
+     ``writer.metadata`` then these lines are automatically filled by Pymagicc.
+   - When ``.SCEN`` files are read, their ``scenario`` metadata is automatically
+     filled from the third line of the input file (similarly, ``scenario`` is also
+     written to the third line of the file when writing ``.SCEN`` files).
+
+#. Files ending in ``.prn`` are emissions input files for ozone
```
lewisjared: Perhaps add something re prn files no longer being supported

pymagicc/magicc_time.py
```diff
@@ -0,0 +1,201 @@
...
+            np.abs(yfrac - other) < self._convert_to_decimal_required_precision
+        )[0]
+        if match_idx.size == 0:
+            return False
+        if must_be_greater and yfrac < other[match_idx]:
+            return False
+        return True
+
+
+MAGICC_TIME_CONVERTER = _MAGICCTimeConverter()
```
lewisjared: Why are you using a class here? purely for caching? Check out `functools.lru_cache`

I wouldn't capitalise MAGICC_TIME_CONVERTER as I tend to reserve that for constants where this is a singleton pattern. Pure stylistic choice and not sure what PEP8 says...

lewisjared: Another solution would be to add the following lines so you can use `from pymagicc.magicc_time import convert_to_decimal_year`

```
convert_to_decimal_year = MAGICC_TIME_CONVERTER.convert_to_decimal_year
convert_to_datetime = MAGICC_TIME_CONVERTER.convert_to_datetime
```

pymagicc/io.py
```diff
@@ -401,12 +400,38 @@ def process_header(self, header):
...
+                else:
+                    if ":" not in line:
+                        header_lines.append(line)
+                    else:
+                        bits = line.split(":")
+                        key = bits[0]
+                        value = ":".join(bits[1:])
+                        metadata[key.strip()] = value.strip()
+
+        if header_lines:
```
lewisjared: So files with no header lines won't have a 'header' key in their metadata. Is that the intended functionality because now you always have to check if the key exists where as a blank header is effectively the same thing

docs/file_conventions.rst
znicholls: it says that already right?

pymagicc/magicc_time.py
znicholls: oops ye I should have used `functools.lru_cache` will change

znicholls:
> Why not reformat all the included files?

Consistency with what is publicly available

pymagicc/io.py
znicholls: The thinking is this: if there’s a header there, use it. If there’s not, don’t add it so the user is forced to add one if/when they re-write. Alternately we could do what we do in `_MAGWriter` and autofill when writing if header isn’t there/is blank

lewisjared:
LGTM. @rgieseke can you have a look at this?

rgieseke:
I find the general treatment of MAGICC's internal 1/12 yearly values converting to `datetime` objects super confusing but otherwise this looks good to me!

PR #253, opened 2019-07-02
Allow out_dynamic_vars parameter to be used

lewisjared:
# Pull request

- [x] Tests added

Doesn't convert out_dynamic_vars to an integer

### DIFF
--- a/pymagicc/core.py
+++ b/pymagicc/core.py
@@ -777,7 +777,9 @@ class MAGICCBase(object):
 
     def _convert_out_config_flags_to_integers(self, config_dict):
         for key, value in config_dict.items():
-            if key.startswith("out") and key != "out_ascii_binary":
+            # Ignore all out_x_vars variables as that should be a list of strings
+            # Note that there is a variable OUT_ALLOWANYDYNAMICVARS which should be in integer
+            if key.startswith("out") and key != "out_ascii_binary" and not key.endswith('_vars'):
                 config_dict[key] = 1 if value else 0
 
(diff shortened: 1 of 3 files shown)


lewisjared:
#254

lewisjared:
It would be nice to have a beta.2 release after this as this is super handy for the api

pymagicc/core.py
```diff
@@ -777,7 +777,9 @@ def _format_config(self, config_dict):
 
     def _convert_out_config_flags_to_integers(self, config_dict):
         for key, value in config_dict.items():
-            if key.startswith("out") and key != "out_ascii_binary":
+            # Ignore all out_x_vars variables as that should be a list of strings
+            # Note that there is a variable OUT_ALLOWANYDYNAMICVARS which should be in integer
```
znicholls: What's `OUT_ALLOWANYDYNAMICVARS`?

znicholls:
> It would be nice to have a beta.2 release after this as this is super handy for the api

Can do

pymagicc/core.py
lewisjared: Malte has added a variable to force the calculation of some of the summary variables. Previously these were only calculated when OUT_KEYDATA_X was set. This isn't in master yet

znicholls: ok so a discussion for the MAGICC repo then

znicholls:
@lewisjared https://pypi.org/project/pymagicc/2.0.0b3/

PR #257, opened 2019-07-05
Suggested changes on stderr

znicholls:


lewisjared:
Great thanks. Note that the stderr will still be captured for magicc6, albeit that its empty unless the run failed. In that case the cryptic message is never actually useful...

PR #259, opened 2019-08-08
Added strict mode for MAGICCBase

lewisjared:
# Pull request

- [x] Tests added

Optional strict parameter to allow runs to continue with invalid configuration

### DIFF
--- a/pymagicc/core.py
+++ b/pymagicc/core.py
@@ -84,7 +84,7 @@ class MAGICCBase(object):
     version = None
     _scen_file_name = "SCENARIO.SCEN7"
 
-    def __init__(self, root_dir=None):
+    def __init__(self, root_dir=None, strict=True):
         """
         Initialise
 
@@ -94,10 +94,15 @@ class MAGICCBase(object):
             Root directory of the MAGICC package. If ``None``, a temporary
             copy of MAGICC is made based on the result of `
             `self.get_exectuable()``.
+        strict: bool
+            If True, enforce the configuration checks, otherwise a warning
+            is raised if any invalid configuration is found and the run is
+            continued.
         """
         self.root_dir = root_dir
         self.config = None
         self.executable = self.get_executable()
+        self.strict = strict
 
         if root_dir is not None:
             self.is_temp = False
@@ -408,6 +413,12 @@ class MAGICCBase(object):
 
         return [f + ".BINOUT" for f in bin_out] + extras
 
+    def _check_failed(self, msg):
+        if self.strict:
+            raise ValueError(msg)
+        else:
+            warnings.warn(msg)
+
     def check_config(self):
         """Check that our MAGICC ``.CFG`` files are set to safely work with PYMAGICC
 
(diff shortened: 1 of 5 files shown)


pymagicc/core.py
```diff
@@ -94,10 +94,15 @@ def __init__(self, root_dir=None):
             Root directory of the MAGICC package. If ``None``, a temporary
             copy of MAGICC is made based on the result of `
             `self.get_exectuable()``.
+        strict: bool
+            If True, enforce the configuration checks, otherwise a warning
+            is raised if any invalid configuration is found and the run is
+            continued.
```
znicholls:
```suggestion
            continued. Setting ``strict=False`` is only recommended for experienced users of MAGICC.
```

znicholls: One thought I had

rgieseke:
What's the usecase? Legacy files?

znicholls:
Reproducing other work flows. Best practice is to not do this, but if
you’re just trying to reproduce someone else’s work then it saves a bunch
of mucking around trying to work out how to make the two workflows match
(which is what you have to do if you don’t have the not strict option).

tests/test_core.py
```diff
@@ -158,6 +158,75 @@ def test_run_failure_confusing_emissions_scenarios(package, invalid_config):
...
+    ],
+)
+def test_check_config_non_strict(package_non_strict, invalid_config):
+    f90nml.write(invalid_config, join(package_non_strict.run_dir, "MAGCFG_USER.CFG"), force=True)
+
+    error_msg = re.escape(
+        "You have more than one `FILE_EMISSCEN_X` flag set. Using more than one "
+        "emissions scenario is hard to debug and unnecessary with Pymagicc's "
+        "dataframe scenario input. Please combine all your scenarios into one "
+        "dataframe with Pymagicc and pandas, then feed this single Dataframe into "
```
rgieseke:
```suggestion
        "DataFrame with Pymagicc and Pandas, then feed this single DataFrame into "
```

tests/test_core.py
```diff
@@ -158,6 +158,75 @@ def test_run_failure_confusing_emissions_scenarios(package, invalid_config):
...
+        },
+    ],
+)
+def test_check_config_non_strict(package_non_strict, invalid_config):
+    f90nml.write(invalid_config, join(package_non_strict.run_dir, "MAGCFG_USER.CFG"), force=True)
+
+    error_msg = re.escape(
+        "You have more than one `FILE_EMISSCEN_X` flag set. Using more than one "
+        "emissions scenario is hard to debug and unnecessary with Pymagicc's "
+        "dataframe scenario input. Please combine all your scenarios into one "
```
rgieseke:
```suggestion
        "DataFrame scenario input. Please combine all your scenarios into one "
```

znicholls:
@lewisjared if you address the suggestions and get the tests passing I'll merge

lewisjared:
Implemented the suggested changes. The failing tests are from the notebooks and the pyam integration which is not from this PR

znicholls:
> The failing tests are from the notebooks and the pyam integration which is not from this PR

Can you hotfix them by just pinning pandas==0.24.2 in this PR?

PR #261, opened 2019-08-18
Update variable mapping

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```


tests/test_definitions.py
```diff
@@ -92,12 +104,23 @@ def test_convert_magicc7_to_openscm_variables(magicc7, openscm):
     [
         ("HFC245FA_EMIS", "Emissions|HFC245fa"),
         ("CO2I_EMIS", "Emissions|CO2|MAGICC Fossil and Industrial"),
-        ("CH4_EMIS", "Emissions|CH4"),
         ("CH3CCL3_EMIS", "Emissions|CH3CCl3"),
         ("CCL4_INVERSE_EMIS", "Inverse Emissions|CCl4"),
+        ("CO2_EMIS", "Emissions|CO2"),
```
lewisjared: Should this be CO2T_EMIS? In the source there is no reference to a CO2_EMIS variable

pymagicc/definitions/__init__.py
```diff
@@ -283,6 +285,8 @@ def get_openscm_replacement(in_var):
             )
         elif variable.endswith("B") and not edge_case_B:
             variable = DATA_HIERARCHY_SEPARATOR.join([variable[:-1], "MAGICC AFOLU"])
+        elif variable.endswith("T") and not edge_case_B:
```
lewisjared:
```suggestion
        elif variable.endswith("T"):
```

znicholls:
> Does it handle `DAT_SURFACE_TEMP_SUBANNUAL` or `*_GWPEMIS`?

`DAT_SURFACE_TEMP_SUBANNUAL` nope but will add. I'd rather not support `*_GWPEMIS`, I think this sort of conversion should be done in post-processing rather than relying on the values hard-coded in MAGICC to be right...

tests/test_definitions.py
znicholls: I think it doesn't actually matter as MAGICC has to have a fossil/afolu split (i.e. you could never prescribe total CO2 emissions). In any case, I have followed the convention for methane. In the output, `CH4T_EMIS` is total methane emissions but in the input of a SCEN7 file, the column header has to be `CH4_EMIS`. Hence the mapping has to be different depending on whether you're going forwards or backwards.

PR #264, opened 2019-09-23
Handle no output

lewisjared:
Allows an empty MAGICCData object to be returned from run (also ran `make black`)

- [x] Tests added


tests/test_core.py
```diff
@@ -1367,3 +1371,20 @@ def run(*args, **kwargs):
...
+@pytest.mark.slow
+def test_magicc7_empty(package):
+    package.set_output_variables()
+    res = package.run(out_dynamic_vars=[])
+
+    assert len(res) == 0
+
+
+@pytest.mark.slow
+def test_magicc7_empty_with_params(package):
```
znicholls:
```suggestion
def test_empty_output_with_params(package):
```

tests/test_core.py
```diff
@@ -1367,3 +1371,20 @@ def run(*args, **kwargs):
                 assert len(record) == 0
     except FileNotFoundError:
         pytest.skip("MAGICC7 not installed")
+
+
+@pytest.mark.slow
+def test_magicc7_empty(package):
```
znicholls:
```suggestion
def test_empty_output(package):
```

znicholls:
Sorry @lewisjared just thinking about this more. Could we have the case of returning empty be a special case (e.g. `only="empty"`)? I'm just thinking that returning empty dataframes is a special case, and for most users it would be better to raise an error message if there's no output. Option b would be that if `not strict` then returning empty works, otherwise it raises the error (in line with `strict` being for advanced users).

lewisjared:
I don't like `only="empty"`, but option b would work

lewisjared:
Added a check for strict

lewisjared:
And appveyor is broken

znicholls:
cool rebase then we're good

PR #266, opened 2019-10-21
Switch to using a whitelist for out_ flags

lewisjared:
# Pull request

- [x] Tests added
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 master
 ------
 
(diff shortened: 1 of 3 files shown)


znicholls:
lgtm. I'd probably prefer to have just added exceptions to the rule rather than the whitelist but that's splitting hairs. If you rebase on #267 this should pass (I think).

lewisjared:
I started on that but realised the list of exceptions are just as long. These shouldn't change in future as `out_dynamic_vars` becomes the norm

znicholls:
cool rebase then we're good

PR #267, opened 2019-10-21
Hotfix appveyor

znicholls:
Attempt to hotfix appveyor so #264 can continue


znicholls:
@lewisjared let's see what happens here

znicholls:
@lewisjared if you're happy (and the tests pass), you should be able to merge this and then rebase #264

PR #277, opened 2019-12-22
Compact out readers

znicholls:
# Pull request

- [x] #281 
- [x] #288 
- [x] #276 
- [x] Tests added

## Adding to CHANGELOG.rst

```
```


znicholls:
@rgieseke and @lewisjared this is good to go I think. I've made a separate issue for docs (#279) so we don't forget but I don't think there's much point doing docs now as MAGICC7 isn't available.

pymagicc/io.py
```diff
@@ -1344,6 +1342,164 @@ def find_region(x):
...
+        ts["variable"] = ts.index.map(
+            lambda x: convert_magicc7_to_openscm_variables(
+                x.split("__")[0].replace("DAT_", "")
+            )
+        )
+        ts["region"] = ts.index.map(
+            lambda x: convert_magicc_to_openscm_regions(x.split("__")[1])
+        )
+        ts["year"] = ts.index.map(lambda x: x.split("__")[2])
+        if not (ts["year"].apply(len) == 4).all():  # pragma: no cover # safety valve
```
lewisjared: This is checking the length of an int cast to a string right? It might be clearer to cast the years to ints in 1 line and then check for uniqueness

```
ts["year"] = ts.index.map(lambda x: x.split("__")[2]).astype(int)
if not ts["year"].is_unique:
    explode
```

pymagicc/io.py
```diff
@@ -1344,6 +1342,164 @@ def find_region(x):
...
+            raise NotImplementedError("Non-annual data not yet supported")
+
+        ts["year"] = ts["year"].astype(int)
+
+        ts = ts.reset_index(drop=True)
+
+        id_cols = {"variable", "region", "year"}
+        run_cols = set(ts.columns) - id_cols
+        ts = ts.melt(value_vars=run_cols, var_name="run_id", id_vars=id_cols)
```
lewisjared: Ahh is this how you reindex using the id cols and adding a run_id col?

pymagicc/io.py
```diff
@@ -1344,6 +1342,164 @@ def find_region(x):
...
+
+        ts["unit"] = "unknown"
```
lewisjared: We should have thought about that before hand.. 

Ohh well this format might be short lived

pymagicc/io.py
```diff
@@ -1344,6 +1342,164 @@ def find_region(x):
...
+
+        new_index = list(set(ts.columns) - {"value"})
+        ts = ts.set_index(new_index)["value"].unstack("year")
+
+        paras = compact_table[para_cols]
+        paras.index.name = "run_id"
+
+        # Find the columns which group together as lists
+        cols_to_merge = []
```
lewisjared: Can we extract this to a util function as it is useful for other things too - also this function is getting hard to reason about. The separator '_' should be able to be specified. Could be put in `scmdata` as a common dep

pymagicc/io.py
znicholls: yep good suggestion

pymagicc/io.py
znicholls: it's how you go from having each run being a column to having run_id be one column and value be one column (it's undoing a level of pivoting)

pymagicc/io.py
znicholls: > We should have thought about that before hand

We did but wouldn't have worked with MCMC stuff etc. etc.

> Ohh well this format might be short lived

It's fine we can always update in future. For now it's just useful to be able to get this into a dataframe easily.

pymagicc/io.py
znicholls: I lie, at this point the years won't be unique cause you'll have e.g. `DAT_CO2_CONC_1765` and `DAT_SURFACE_TEMP_1765`.

pymagicc/io.py
znicholls: > Could be put in `scmdata` as a common dep

Surely this is very MAGICC specific? I've done some refactoring, there's not really sensible way to do this though cause of MAGICC's config (see the tests at the end of `tests/test_io.py`).

znicholls: spoke in person, make super MAGICC specific so it ignores e.g. tuningmodel, emisscen and keydata

znicholls: resolved by ec508e4

PR #283, opened 2020-03-04
Add diagnose tcre

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

Depends on #282


pymagicc/core.py
```diff
@@ -1258,17 +1287,17 @@ def default_config(self):
 
         return self._default_config
 
-    def _check_tcr_ecs_total_RF(self, df_total_rf, tcr_time, ecs_time):
-        super()._check_tcr_ecs_total_RF(df_total_rf, tcr_time, ecs_time)
+    def _check_tcr_ecs_tcre_total_RF(self, df_total_rf, tcr_time, ecs_time):
```
lewisjared: Why perform extra checks for the subclass. Is there a real world example which passes the Base classes implementation, but fails this check?

tests/test_core.py
```diff
@@ -548,74 +579,80 @@ def assert_bad_tcr_ecs_diagnosis_values_caught(
...
+        np.testing.assert_allclose(actual_result["ecs"], 2.9968448)
+        np.testing.assert_allclose(actual_result["tcre"], 0.00229016699)
+
     if isinstance(package, MAGICC7):
         # see how stable this is, can delete the test later if it's overly restrictive
-        assert actual_result["tcr"] == 2.0875821584565704
-        assert actual_result["ecs"] == 3.155772857010324
+        np.testing.assert_allclose(actual_result["tcr"], 1.982697)
+        np.testing.assert_allclose(actual_result["ecs"], 2.9948422)
+        np.testing.assert_allclose(actual_result["tcre"], 0.0023189736)
```
lewisjared: That's a 1% difference for TCRE between M6 and M7. Is that just differences in defaults

pymagicc/core.py
znicholls: Yes, running these checks with MAGICC7 doesn't work because it's CO2 RF isn't always logarithmic.

tests/test_core.py
znicholls: Yep and new ocean parameterisation etc.

znicholls:
@lewisjared , do you want me to update those lines before you merge or do the replies make sense?

lewisjared:
Makes sense

PR #290, opened 2020-04-24
Use openscm_units

lewisjared:
# Pull request

Bumping scmdata to 0.4.3


setup.py
```diff
@@ -45,8 +45,8 @@
     "pandas-datapackage-reader",
     "f90nml",
     "PyYAML",
-    "scmdata>=0.4",
-    "scipy",
+    "scmdata>=0.4.3",
+    "scipy"
```
znicholls, line 49:
```suggestion
```

think we could bin this...

lewisjared:
why doesn't wine32 install wine in /usr/bin, but rather /usr/lib/wine...

lewisjared:
Now all the plotting is broken???

PR #292, opened 2020-04-28
Update DAT ERF handling

znicholls:


znicholls:
this should be right now @lewisjared

PR #299, opened 2020-06-26
Undo fortran cases

znicholls:
# Pull request

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

### DIFF
--- a/pymagicc/io.py
+++ b/pymagicc/io.py
@@ -337,7 +337,7 @@ class _Reader(object):
     def _get_column_headers_and_update_metadata(self, stream, metadata):
         if self._magicc7_style_header():
             column_headers, metadata = self._read_magicc7_style_header(stream, metadata)
-            column_headers = self._read_units(column_headers)
+
         else:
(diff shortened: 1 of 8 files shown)


pymagicc/io.py
```diff
@@ -2679,7 +2687,7 @@ def _write_variable_datablock_radforcing(self, output, data_block, units_level):
             data_block, units_level
         )
 
-        units = [u.replace("Wpermsuper2", "W/m2") for u in units]
+        units = [u.replace("W_per_msuper2", "W/m2") for u in units]
```
lewisjared, line 2690: Why `super` not `_super_` for consistency?

znicholls: I'm using "_" for space at the moment and I don't think pint would work if we had "W m ^ 2"

znicholls: although maybe it would...

lewisjared: Ahh makes sense. It thought it was to tokenise it

PR #303, opened 2020-07-28
pymagicc.io refactor

lewisjared:
# Pull request

Please confirm that this pull request has done the following:

- [x] Description in ``CHANGELOG.rst`` added

Separated `pymagicc.io` from a module into a package. Each file type is now in its own module to make it easier to edit in future.

A future cleanup would be splitting the tests, but I tried to not touch them in the PR

PR #305, opened 2020-07-29
Update bin reader

lewisjared:
# Pull request

Requires
- [x] #303 

- [x] Tests added


pymagicc/io/binout.py
```diff
@@ -147,4 +139,118 @@ def process_data(self, stream, metadata):
             "todo": ["SET"] * len(regions),
```
lewisjared, line 139: @znicholls Should the dat files be SET or N/A by default? The ASCII files are N/A and the old bin reader used SET.

znicholls: Hmm good question. I would say N/A and change the old bin reader

znicholls:
sorry to be a pain but can we use e.g. "not_relevant" or just not include "TODO" at all (probably easier to use "not_relevant" for now for consistency with the ASCII, or also remove TODO from the OUT readers). Having "N/A" in metadata causes all sorts of issues with groupby etc. because pandas thinks it's nan.

lewisjared:
> sorry to be a pain but can we use e.g. "not_relevant" or just not include "TODO" at all (probably easier to use "not_relevant" for now for consistency with the ASCII, or also remove TODO from the ASCII OUT readers). Having "N/A" in metadata causes all sorts of issues with groupby etc. because pandas thinks it's nan.

Had a quick look at getting rid of the TODO column for the out/binout readers, but it ended up being a reasonable change so let's leave that for another day. Changing the out and the binout reader default TODO value to "not_relevant"

PR #308, opened 2020-07-31
Unpin pandas and use new pip features

znicholls:
I think the install failures are a pip issue not our scmdata packaging...


lewisjared:
Ahh didn't know about that feature. I would have assumed that it follows the children deps, but I was wrong

znicholls:
ye it's bizarre that this wasn't the default, anyway

znicholls:
yore

PR #311, opened 2020-10-02
Fix naming

znicholls:
# Pull request

Fixes some of the naming to bring us closer to RCMIP conventions

- [x] Tests added

## Adding to CHANGELOG.rst

```
```

### DIFF
--- a/.github/workflows/ci-cd-workflow.yml
+++ b/.github/workflows/ci-cd-workflow.yml
@@ -26,7 +26,7 @@ jobs:
       run: |
         black --check pymagicc tests setup.py --exclude pymagicc/_version.py
         black-nb --check notebooks
-        isort --check-only --quiet --recursive pymagicc tests setup.py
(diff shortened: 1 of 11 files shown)


znicholls:
@lewisjared sorry about the black changes included here...

I wonder if now is the time to bite the bullet and switch things across to RCMIP names, or do we think that will cause too much chaos in downstream uses?

znicholls:
> I wonder if now is the time to bite the bullet and switch things across to RCMIP names, or do we think that will cause too much chaos in downstream uses?

Actually let's not do it everywhere now, chaos really would ensue. One for pymagicc 3.0, can be tracked with in #312.

lewisjared (approved):
LGTM. One decision left:
* 'Surface Temperature' to 'Surface Air Temperature Change'
I'd keep it out of this PR and deal with that pain later. It isn't 100% RCMIP, but close enough...

PR #327, opened 2021-07-02
Ignore `append` docstring

lewisjared:
This comes from scmdata, updating scmdata caused this cell to fail

Once merged I will release v2.1.1


znicholls (approved):
Could also not ignore (I don't hate updating the notebooks when this happens., it's not a bad reminder..)

znicholls:
> Once merged I will release v2.1.1

Nice

PR #328, opened 2021-07-15
Adds additional spelling options for YEARS header

gabriel-abrahao:
# Pull request

Hi openscm people. We at PIK are considering using some of the stable openscm tools in our workflows. For that, it would be super useful if pymagicc could just use our current MAGICC scenario files. 

The only thing stopping that from happening is a line in `read_data_block` that calls `_read_data_header_line` enforcing that the header for the Years variable is spelled in uppercase:

`variables = self._read_data_header_line(self._stream, ["YEARS", "YEAR"])`

This PR currently simply adds a mixed-case "Years" entry to that list, since that's how our models write scenario files write that header:

`variables = self._read_data_header_line(self._stream, ["Years","YEARS", "YEAR"])`

We could change this in our models, but this would be great for backwards compatibility and MAGICC (at least our binary) doesn't seem to care either way. Would you mind merging this? Also, If you'd prefer a more general way to do this, I'd be happy to implement it.

Thanks,
Gabriel


znicholls:
Hi @gabriel-abrahao, thanks for the PR. The changes look fine. Could you please add a test and a CHANGELOG entry?

gabriel-abrahao:
Thanks! I added a test REMIND scenario file and a `test_load_remind_scen` to read it and assert some values. It's passing on my side. Also added an entry to the changelog under `master`, not sure if that's where it should go.

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ The changes listed in this file are categorised as follows:
 master
 ------
 
+Added
+~~~~~
+
+- (`#328 <https://github.com/openscm/pymagicc/pull/328>`_) Adds additional spelling options for YEARS header
```
znicholls, line 23:
```suggestion
- (`#328 <https://github.com/openscm/pymagicc/pull/328>`_) Adds additional spelling options for YEARS header in `.SCEN` files
```

znicholls (commented):
One suggested tweak, otherwise looks good to me thanks!

CHANGELOG.rst
gabriel-abrahao: Nice, that's better!

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ The changes listed in this file are categorised as follows:
 master
 ------
 
+Added
+~~~~~
+
+- (`#328 <https://github.com/openscm/pymagicc/pull/328>`_) Adds additional spelling options for YEARS header in `.SCEN` files
```
znicholls, line 23:
```suggestion
- (`#328 <https://github.com/openscm/pymagicc/pull/328>`_) Adds additional spelling options for YEARS header in ``.SCEN`` files
```

Sorry I got the rst syntax wrong, can you add this then I'll merge

znicholls:
released in v2.1.1 https://pypi.org/project/pymagicc/2.1.1/

PR #330, opened 2021-07-30
Add citation cff file

rgieseke:
With GitHub adding support for CITATION.cff files to display a citation info on the side (see e.g. https://github.com/IAMconsortium/pyam/) i though it might be worth adding one here too.

https://twitter.com/natfriedman/status/1420122675813441540

With Zenodo also using this for displaying info on their website when dong releases i'm not sure how all of this will render out.

https://twitter.com/zotero/status/1420515377390530560

I've copied the citation info from the Readme and added a link to the 'latest' DOI from Zenodo.

We also ought to create releases to trigger a Zenodo archival, not sure if some beta versions should also be archived as they might have been used in papers?

Full spec: https://github.com/citation-file-format/citation-file-format

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -2,7 +2,7 @@ cff-version: 1.1.0
 message: |
   If you make any use of MAGICC, its license requires citing of
 
-    M. Meinshausen, S. C. B. Raper and T. M. L. Wigley (2011). "Emulating coupled atmosphere-ocean and carbon cycle models with a simpler model, MAGICC6: Part I "Model Description and Calibration." Atmospheric Chemistry and Physics 11: 1417-1456. https://doi.org/10.5194/acp-11-1417-2011
+    M. Meinshausen, S. C. B. Raper and T. M. L. Wigley (2011). "Emulating coupled atmosphere-ocean and carbon cycle models with a simpler model, MAGICC6: Part 1 "Model Description and Calibration." Atmospheric Chemistry and Physics 11: 1417-1456. https://doi.org/10.5194/acp-11-1417-2011
 
   If you use Pymagicc in your research, please additionally cite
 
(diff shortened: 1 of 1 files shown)


rgieseke:
One could potentially also have an update of the DOI field included in the CI automation.

znicholls:
Nice. There's a chicken and egg issue with zotero (zotero reads the cff, but the cff can't be properly finalised until the new zotero DOI is minted, which requires the cff etc.). I don't know if new zotero DOIs are somehow predictable. Using latest the whole time is probably also ok.

We can probably make things tidier by using the reference key (https://github.com/citation-file-format/citation-file-format/blob/main/README.md#references-optional) rather than putting all the citations in the message. 

This also allows a citation file to be generated from a zenodo entry: https://github.com/citation-file-format/doi2cff. I don't know if that helps with the chicken and egg problem though.

Maybe the workflow would be (although it's fiddly to automate) 

1. tag release
2. reserve zenodo DOI (there are ways to automate this I think)
3. update citation CFF with new zenodo DOI and version
4. push to zenodo and publish release

https://github.com/openscm/openscm-zenodo should be able to handle all the zenodo stuff (might need an upgrade to allow automatic publishing though).

rgieseke:
Yeah, i looked into the 'references' field but couldn't find a tool quickly which would convert citations to cff format :-]
I was also unsure how GitHub would render these and the initial message.

Good point, on the DOI minting, i think it's maybe not worth the trouble ... Probably easiest to just use the 'latest' DOI or skip it altogether.
I think the main point of the Zenodo integration is that it uses it to read and format the title and authors (so no need for a `zenodo.json` file or manually editing those values on Zenodo.

rgieseke:
I'll add the references and keywords and then we can check how it looks?

znicholls:
:+1:

rgieseke:
>> Presumably I'll have to specify the version tag in the cff for GitHub's widget to work. But Zenodo will get the git tag via the hook data, right? Feels a bit backwards.

> Zenodo will use the version tag in the CFF instead of the hook data.

https://twitter.com/xrotwang/status/1420389754563145730

CITATION.cff
```diff
@@ -0,0 +1,66 @@
...
+        given-names: M.
+      - family-names: Raper
+        given-names: S. C. B.
+      - family-names: Wigley
+        given-names: T. M. L.
+    title: "Emulating coupled atmosphere-ocean and carbon cycle models with a simpler model, MAGICC6: Part 1: Model Description and Calibration"
+    date-published: 2011-02-16
+    doi: 10.5194/acp-11-1417-2011
+  - type: article
+    authors:
```
znicholls, line 57: I'd probably add the journal here too, same above

znicholls:
Otherwise looks good

znicholls:
I think this is ready to go once #331 is merged and you rebase @rgieseke

znicholls (approved):

